PR #375, opened 2013-03-30
Add --utc to graph option.

jyavenard:
This allow to easily calculate daily average ranging from midgnight to midnight rather that say the last 24 hours.


netniV:
This patch is seven years old! :)  

But I like the idea, if I was to target this towards the current version, would you commit it?  Will have to try to co-author it as @jyavenard did the original work.

jyavenard:
FWIW, I still use it daily, and I've been updating it whenever a new version of rrdtool came out

netniV:
Maybe the better course of action then is to close and recreate the PR 👍   I'll chip my vote in for it when it's reposted if you tag me

oetiker:
"fixing" this one is fine

PR #376, opened 2013-03-30
Add --utc switch

jyavenard:
This allow to easily calculate daily average ranging from midgnight to midnight rather that say the last 24 hours.


jyavenard:
seems I had to pull request opened, remove one

PR #579, opened 2015-01-29
Fix flock being always disabled

siboulet:
Typo in configure script causing flock to always be disabled.
Introduced in 0033ee5


oetiker:
thanks

PR #582, opened 2015-01-31
Add support for storing RRD in Ceph through librados

siboulet:
This patch adds support for creating, updating and retrieving RRD files on a Ceph cluster. It adds a ceph// prefix to RRD file name (similar to sql// used by libdbi). It uses librados to talk natively to Ceph.

Ceph support is compiled when librados is found at compile time. A --disable-librados flag was also added. 

The interfaces provided by the callback (cb//) or libdbi methods were not sufficient as I wanted to write / update RRD through librrd, and also wanted to preserve the RRD file format.

Tested create, dump, update and fetch operations.

I'd really like to see this merged. Is this something you would consider merging? Please review and let me know any changes you'd like to see. Also need to look at updating the documentation.

Thanks

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -542,6 +542,18 @@ AC_ARG_ENABLE(libdbi,AS_HELP_STRING([--disable-libdbi],[do not build in support 
 
 AM_CONDITIONAL(BUILD_LIBDBI,[test $have_libdbi != no])
 
+have_librados=no
+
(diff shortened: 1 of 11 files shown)


oetiker:
@siboulet neat ... I do not use ceph myself. can you elaborate  on the advantages of running rrdtool with ceph ? I am inclined to merge, but your patch does not include documentation yet.

siboulet:
Added documentation

oetiker:
thanks

PR #586, opened 2015-02-26
Fix I/O error detection in dump/restore

mkhon:
The main reason for the fixes was 'rrdtool dump' not detecting I/O errors (exit code 0) and leaving incompletely written .rrd files in that case.

I also fixed a number of other bugs like:
- free'ing unallocated memory
- missing rrd_free() call

### DIFF
--- a/src/rrd_create.c
+++ b/src/rrd_create.c
@@ -1437,14 +1437,21 @@ int write_fh(
         /* we output 3 or higher */
         strcpy(rrd->stat_head->version, "0003");
     }
-    fwrite(rrd->stat_head, sizeof(stat_head_t), 1, fh);
-    fwrite(rrd->ds_def, sizeof(ds_def_t), rrd->stat_head->ds_cnt, fh);
-    fwrite(rrd->rra_def, sizeof(rra_def_t), rrd->stat_head->rra_cnt, fh);
-    fwrite(rrd->live_head, sizeof(live_head_t), 1, fh);
-    fwrite(rrd->pdp_prep, sizeof(pdp_prep_t), rrd->stat_head->ds_cnt, fh);
-    fwrite(rrd->cdp_prep, sizeof(cdp_prep_t),
+
+#define FWRITE_CHECK(ptr, size, nitems, fp)	               \
+    do {                                                       \
+        if (fwrite((ptr), (size), (nitems), (fp)) != (nitems)) \
+            return (-1);                                       \
+    } while (0)
+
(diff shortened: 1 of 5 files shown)


src/rrd_create.c
```diff
@@ -1437,14 +1437,21 @@ int write_fh(
...
-    fwrite(rrd->rra_def, sizeof(rra_def_t), rrd->stat_head->rra_cnt, fh);
-    fwrite(rrd->live_head, sizeof(live_head_t), 1, fh);
-    fwrite(rrd->pdp_prep, sizeof(pdp_prep_t), rrd->stat_head->ds_cnt, fh);
-    fwrite(rrd->cdp_prep, sizeof(cdp_prep_t),
+
+#define FWRITE_CHECK(ptr, size, nitems, fp)                    \
+    do {                                                       \
+        if (fwrite((ptr), (size), (nitems), (fp)) != (nitems)) \
+            return (-1);                                       \
+    } while (0)
```
oetiker: a loop which exits after the first round anyway ? what is the effect of this ?

src/rrd_dump.c
```diff
@@ -81,7 +81,12 @@ int rrd_dump_cb_r(
 //These two macros are local defines to clean up visible code from its redndancy
 //and make it easier to read.
 #define CB_PUTS(str)                                            \
-    cb((str), strlen((str)), user)
+    do {							\
+        size_t len = strlen(str);				\
+								\
+        if (cb((str), len, user) != len)                        \
+            goto err_out;                                       \
+    } while (0);
```
oetiker: here I understand .... local context for the lan variable

src/rrd_create.c
mkhon: 1) this is to protect against the following:

if (...)
   FWRITE_CHECK(...);
else
   /* do something */

Without do/while 'else' will be related to the 'if' inside the FWRITE_CHECK (and this is not what the user of this macro is expected).

2) It is generally a good practice to wrap macros that do not expand to expressions (like this one, or any multi-statement macros) so that the user of macro can use it as a single statement (as in the example above). do/while(0) will anyway be optimized out but it turns a multi-statement macro to a "single-statement".

src/rrd_dump.c
mkhon: Not only the context but a wrapper for if(..) as in the previous case.

src/rrd_create.c
oetiker: ok, cool :)

oetiker:
thanks

PR #594, opened 2015-03-15
Add some value pretty-printing depending on what data type is graphed

etbuira:


oetiker:
Can you provide some use case example of your proposed changes ?

etbuira:
Hi,

Admitedly, most of the graphs won't use those features.

Rendering timestamps is interesting when graphing difference with 'sampling' time or now() would loose some of the meaning. Beyond my specific need, a possibly interesting use case is graphing expected finish time in project management (and how it evolves over time).

About durations formatting, it gets interesing as soon as expected duration can span a large scale. A use case is representing time of presence of elements in a work queue that is processed both automatically for most common cases and by humans for other cases.

When making those patches, i've seen those pretty printings a bit like image_desc_t.base, that simplifies the patches. But i admit it can be nice to scale/offset those PPs with some other data. I'll do the change if you find it interesting enough to consider merging.

Regards.

oetiker:
sure I'd like to merge, I just want to make sure it properly fits the logic of rrdtool graphs logic

etbuira:
Fair enough, closing this merge request for a new one

PR #595, opened 2015-03-16
Add some value formatting depending on what data type is graphed

etbuira:
Hi,

This new version allows to format each element independently of the others.

Regards.


oetiker:
thank you!

etbuira:
Thank you.

PR #598, opened 2015-03-23
Added resolution (step) option to rrdcached FETCH(BIN) commands.

mindnever:
This patch adds -r option to the FETCH and FETCHBIN commands.


oetiker:
nice ... 

please update src/rrd_client.c, src/rrd_fetch.c and doc/rrdcached.pod accordingly.

oetiker:
ping

mindnever:
Sorry for not acking. Will do, just had some other things on my mind :)

oetiker:
I'm afraid this will not make it into 1.5 like that

PR #601, opened 2015-04-01
Fix pango fontmap threading issue.

### LINKED ISSUE #600: Fails to graph with threaded calls
This seems to be related to https://github.com/oetiker/rrdtool-1.x/issues/386 and https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=758868 .

When trying to graph in parallel (different threads) it ends up in an infinite loop generating the following output:
```
(process:1): GLib-GObject-CRITICAL **: g_object_get_qdata: assertion 'G_IS_OBJECT (object)' failed
(process:1): GLib-GObject-CRITICAL **: g_object_replace_qdata: assertion 'G_IS_OBJECT (object)' failed
(process:1): GLib-GObject-CRITICAL **: g_object_get_qdata: assertion 'G_IS_OBJECT (object)' failed
(process:1): GLib-GObject-CRITICAL **: g_object_replace_qdata: assertion 'G_IS_OBJECT (object)' failed
(process:1): GLib-GObject-CRITICAL **: g_object_get_qdata: assertion 'G_IS_OBJECT (object)' failed
(process:1): GLib-GObject-CRITICAL **: g_object_replace_qdata: assertion 'G_IS_OBJECT (object)' failed
```

I have been running the following test program to reproduce the problem. Replacing the last line with `thr.run()` works as expected.
```python
import rrdtool
import time
import threading

rrdtool.create(
    '/tmp/myrrd',
    "--step", "1800", "--start", '0',
    "DS:metric1:GAUGE:2000:U:U",
    "DS:metric2:GAUGE:2000:U:U",
    "RRA:AVERAGE:0.5:1:600",
    "RRA:AVERAGE:0.5:6:700",
    "RRA:AVERAGE:0.5:24:775",
    "RRA:AVERAGE:0.5:288:797",
    "RRA:MAX:0.5:1:600",
    "RRA:MAX:0.5:6:700",
    "RRA:MAX:0.5:24:775",
    "RRA:MAX:0.5:444:797"
)

t = int(time.time())
for i in xrange(30):
    rrdtool.update('/tmp/myrrd', '%s:%s:%s' % (t + (i * 1800), i, 3))


class GraphThread(threading.Thread):
    def __init__(self, i, t, *args, **kwargs):
        self.i = i
        self.t = t
        super(GraphThread, self).__init__(*args, **kwargs)

    def run(self):
        rrdtool.graph(
            "myrrd%s.png" % self.i,
            "--start", "%s" % self.t, "--end", "%s" % (self.t + (30*1800)),
            "--vertical-label=Num",
            "-w 800",
            "DEF:m1_num=/tmp/myrrd:metric1:AVERAGE",
            "DEF:m2_num=/tmp/myrrd:metric2:AVERAGE",
            "LINE1:m1_num#0000FF:metric1\r",
            "LINE2:m2_num#00FF00:metric2\r",
        )

for i in xrange(10):
    thr = GraphThread(i, t)
    thr.start()
```

Environment:
 - Debian jessie with RRDTool 1.4.8 (debian version)
 - Debian jessie with RRDTool 1.4.9
 - Debian jessie with RRDTool built from master

Dockerfile used for running with master:
```
# docker build --rm -t test .
# docker run -it test python /testfile.py
FROM debian:jessie
MAINTAINER Mattias Lundberg <lundberg.mattias@gmail.com>

ENV LANG en_US.UTF-8

RUN apt-get update && \
        apt-get install -y git tcl-dev dh-autoreconf quilt libxml2-dev libcairo2-dev libpango1.0-dev  liblua5.1-0-dev groff libdbi0-dev python-all-dev python-all-dbg gem2deb lua5.1 libpcre3-dev && \
        apt-get clean

RUN git clone https://github.com/oetiker/rrdtool-1.x.git rrdtool
RUN cd rrdtool && ./configure && make && make install && make site-python-install

ADD ./testfile.py /testfile.py
```

Running the same on updated Arch Linux (RRDTool 1.4.9) gives the error (non-threaded works):
```
(process:2449): GLib-GObject-CRITICAL **: g_object_ref: assertion 'object->ref_count > 0' failed
(process:2449): Pango-CRITICAL **: pango_font_map_get_serial: assertion 'PANGO_IS_FONT_MAP (fontmap)' failed
(process:2449): Pango-CRITICAL **: pango_font_map_get_serial: assertion 'PANGO_IS_FONT_MAP (fontmap)' failed
(process:2449): Pango-CRITICAL **: pango_font_map_get_serial: assertion 'PANGO_IS_FONT_MAP (fontmap)' failed
zsh: segmentation fault (core dumped)  python2 rrd_tool_test.py
```

JoakimSoderberg:
Since pango v1.32.6 `pango_cairo_font_map_get_default` returns a per thread `fontmap`. But since we're allocating this statically on the first call to `rrd_init`,
if we then try to access it again via another thread glib will get stuck in an infinte assertion loop.

Instead use `pango_cairo_font_map_new()` when creating the `fontmap`. This will not create a per thread version and glib will be happy.

Also added some extra locking whenever using the `im->layout` pango structure that uses the `fontmap` internally as a precaution.

This fixes #600

### DIFF
--- a/src/rrd_graph.c
+++ b/src/rrd_graph.c
@@ -461,9 +461,11 @@ int im_free(
         free(im->rendered_image);
     }
 
+    mutex_lock(im->fontmap_mutex);
     if (im->layout) {
         g_object_unref(im->layout);
     }
+    mutex_unlock(im->fontmap_mutex);
 
 	if (im->ylegend)
 		free(im->ylegend);
@@ -4656,6 +4658,7 @@ void rrd_graph_init(
     unsigned int i;
     char     *deffont = getenv("RRD_DEFAULT_FONT");
     static PangoFontMap *fontmap = NULL;
+    static mutex_t fontmap_mutex = MUTEX_INITIALIZER;
     PangoContext *context;
 
     /* zero the whole structure first */
@@ -4740,6 +4743,7 @@ void rrd_graph_init(
 
     im->surface = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, 10, 10);
     im->cr = cairo_create(im->surface);
+    im->fontmap_mutex = &fontmap_mutex;
 
     for (i = 0; i < DIM(text_prop); i++) {
         im->text_prop[i].size = -1;
@@ -4747,8 +4751,10 @@ void rrd_graph_init(
         rrd_set_font_desc(im,i, deffont ? deffont : text_prop[i].font,text_prop[i].size);
     }
 
+    mutex_lock(im->fontmap_mutex);
+
     if (fontmap == NULL){
-        fontmap = pango_cairo_font_map_get_default();
+        fontmap = pango_cairo_font_map_new();
     }
 
 #ifdef HAVE_PANGO_FONT_MAP_CREATE_CONTEXT
@@ -4772,7 +4778,7 @@ void rrd_graph_init(
         (im->font_options, CAIRO_HINT_METRICS_ON);
     cairo_font_options_set_antialias(im->font_options, CAIRO_ANTIALIAS_GRAY);
 
-
+    mutex_unlock(im->fontmap_mutex);
 
     for (i = 0; i < DIM(graph_col); i++)
         im->graph_col[i] = graph_col[i];
@@ -5369,8 +5375,10 @@ void rrd_graph_options(
         }
     } /* while (1) */
 
+    mutex_lock(im->fontmap_mutex);
     pango_cairo_context_set_font_options(pango_layout_get_context(im->layout), im->font_options);
     pango_layout_context_changed(im->layout);
+    mutex_unlock(im->fontmap_mutex);
 
 
     if (im->primary_axis_format != NULL && im->primary_axis_format[0] != '\0') {
--- a/src/rrd_graph.h
+++ b/src/rrd_graph.h
@@ -25,6 +25,7 @@
 
 #include "rrd_tool.h"
 #include "rrd_rpncalc.h"
+#include "mutex.h"
 
 #include <glib.h>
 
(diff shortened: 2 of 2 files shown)


oetiker:
* what is performance impact of this approach in the single threaded case ?
* at the moment librrdtool comes in two flavours, threaded and non threaded ... the major stumbling block apart from the one you just tackled is our use of getopt ... using popt might be an option for getting rid of this.
* please make sure the platform support the mutex lock by adding appropriate tests to configure.ac

JoakimSoderberg:
Do you have any benchmarking script to check this, or do I have to roll my own?

I mean, I only added the locking because you mentioned it. I have not seen any issues related to it. I guess the fontmap isn't written to but simply read from by multiple threads? However I'm not familiar with all the glib reference counting, I guess that might create issues?

I'm not sure what you mean with the `configure.ac` comment. I simply used the built-in `mutex_t` wrapper. It is used in other places in the code. So what type of check is needed that isn't already done?

Regarding popt you could also checkout the command line parsing lib I've written if you're interested (https://github.com/JoakimSoderberg/cargo). 

Replacing the entire use of getopt is a bit more than I will be able to do. This is basically a 1 line fix for an issue that causes a complete crash. The locking is just an extra precaution.

oetiker:
the locking was just a guess ... no idea if it is necessary ... if not, all  the better ... 

about the multithreading thing ... thats from back in the day when not all platforms supported pthreads ... but I just realized that rrd_client.c is using mutex_t as well and it is even included in the non-threaded version of rrdtool.

I have no benchmarking code as such ... when I initially came up with this fontmap thing, I just create a loop to create a bunch of charts ... 

And cargo ... didn't know about that one ...

JoakimSoderberg:
Thanks for merging :)

PR #602, opened 2015-04-04
reenable GRAD

cheese1:
The original patch from 2009 (i think) has modified eg the AREA part and will likely need some cleanup. GRAD seems to work now. 
Examples are:
http://statistik.nosuchhost.net/hum1week.png
http://statistik.nosuchhost.net/wind1week.png

GRAD seems to have troubles with negative values, not yet clear how to fix that.


oetiker:
Hi Cheese1,

Looking at this again, I think it would make more sense to integrate Gradients with the normal AREA command, since we now have the fancy key=value parser this should be pretty simple.

By just adding two new optional arguments to the AREA command:  gradient-color=#xxkdd:gradient-height=x (or something like that).

As for the spaces/tabs changes, thats is nice as well, but please make it a separate pull request ... reviewing your changes with tons of space/tab changes in between is a bit hard.

oetiker:
I have done the merge into AREA. Thanks for your help!

PR #612, opened 2015-04-25
Straighten command line options

benoit-monin:
This patchset tries to bring some coherence between the short and long command-line options, the documentation and the rrdtool help. It adds code for documented but unimplemented options and it documents the "hidden" options.

After trying to get _rrdtool create -t file.rrd -r file.rrd ..._ working, I took a look at the code only to find that the short options for template and source were not implemented.

As a suggestion, and since rrdtool already relies on glib, it might be a good idea to use its [commandline option parser](https://developer.gnome.org/glib/stable/glib-Commandline-option-parser.html) ?

### DIFF
--- a/doc/rrdcgi.pod
+++ b/doc/rrdcgi.pod
@@ -4,7 +4,7 @@ rrdcgi - Create web pages containing RRD graphs based on templates
 
 =head1 SYNOPSIS
 
-C<#!/path/to/>B<rrdcgi> S<[B<--filter>]>
+C<#!/path/to/>B<rrdcgi> S<[B<--filter>|B<-f>]>
 
 =head1 DESCRIPTION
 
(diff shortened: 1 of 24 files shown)


oetiker:
Cool. thanks ... could you make this patch against the 1.5 branch please.

A switch to with glib argument parser would be great ... that would be something for master ... It is important that we do not break the existing syntax, but if this is possible I would be ever so glad to loose getopt ... are you offering to do this ?

benoit-monin:
Ok, I'll send a separate pull request against 1.5.

I'll see what I can do about converting options parsing to glib, but don't hold your breath ;) Meanwhile, you can still merge this patchset on master.

oetiker:
will later merge the 1.5 changes into master ...

PR #614, opened 2015-04-26
Straighten options for 1.5

benoit-monin:
This is the same patchset as #612, but against branch 1.5.


oetiker:
cool thanks ... for now all bugfixes go into 1.5 ... I will merge 1.5 changes to master in a week or two.

PR #628, opened 2015-07-03
Add permissions to the sockets acquired from systemd during the socke…

manover:
When rrdcached is started via socket activation process, the file descriptors received from systemd never get permissions added to them. Permissions are usually assigned inside read_options() function, but in case of socket activation, we discard all the file descriptors processed there.
As a result, any request gets denied:
```
info /var/lib/munin/switches/home/dlink-snmp_dlink_if_3-send-d.rrd
-1 Permission denied.
```
This PR makes sure permissions get assigned when rrdcached initializes the sockets during open_listen_sockets_systemd() function call.


oetiker:
thanks

PR #639, opened 2015-07-27
Tests: also truncate output to 9 digits if <1

nirgal:
Previous fix worked for XXXXXXXXXe+X numbers.
We also need XXXXXXXXXe-X.


nirgal:
After that, the test suite works on both amd64 and armel. :)

oetiker:
thanks

PR #641, opened 2015-07-28
Python example

nirgal:
When using python rrdtool.graph(), the locale environment is ignored and months names were always output in English.

The first patch add a hint in documentation `pydoc rrdtool.graph` about calling setlocale first.

The second patch add the stripes example, re-implemented in python (with the locale.setlocale call).

### DIFF
--- a/examples/Makefile.am
+++ b/examples/Makefile.am
@@ -10,7 +10,8 @@ EXTRA_DIST = cgi-demo.cgi.in
 
 examplesdir = $(pkgdatadir)/examples
 examples_SCRIPTS = cgi-demo.cgi piped-demo.pl shared-demo.pl \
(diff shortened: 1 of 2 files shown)


oetiker:
This  is not something python specific but the way locales are to be handled ... see 'man setlocale'

On startup of the main program, the portable "C" locale is selected as default.  A program may be made portable to all locales by calling:

    setlocale(LC_ALL, "");

so mentioning this explicitly for the python bindings seems a bit odd.

Like the example though ...

nirgal:
I forgot the setlocale in the past and it took me forever to understand. But I agree this is not the best place to document that.
I think with the example, this is enough.
So I changed by branch to skip the python docstring change.
It now only have the python example. :)

oetiker:
thanks!

PR #642, opened 2015-08-02
Use 9 digits output for test xport1

nirgal:
This makes the xport1 test work on arm, but breaks it on 10-digits output.
We'll have to find a better trick than DIFF9, because truncating doesn't work when the 10th digit is in [6-9]...
WARNING: merging this as-is will break xport1 unit test on most platforms!


oetiker:
aaah ... DIFF9 should be smarter :)

nirgal:
After last update to DIFF9 in branch 1.5, that patch looks to go. :)

You could also change Makefile.am to add xport1 in TESTS, and add xport1.json.output xport1.xml.output to EXTRA_DIST.

oetiker:
thanks

PR #643, opened 2015-08-02
Hurd first part

nirgal:
Here's the first part of support for systems that don't hard code a file name length limit.

This is based on a patch by Marc Dequènes back in 2008, for version 1.3.7

The general idea for rrd_graph is to have image_desc_t.graphfile as a char*
I wrote a macro RRD_IMGDESC_HAS_FILE(image_desc) because there was a lot of #ifdef in the code. Not sure about the name nor the location, fell free to change them.

A few notes:
- In im_free(), in free(im->graphfile), I'm counting on the fact that free(NULL) is OK.
- In graph_cairo_finish, in error handling, I'm counting on the fact that printf("%s", NULL) prints "(null)". I'm unsure this is portable. Tested ok here.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -251,7 +251,7 @@ AC_C_BIGENDIAN
 dnl for each function found we get a definition in config.h 
 dnl of the form HAVE_FUNCTION
 
-AC_CHECK_FUNCS(nl_langinfo tzset fsync mbstowcs opendir readdir chdir chroot getgid getuid setgid setuid strndup strerror snprintf vsnprintf vasprintf fpclass class fp_class isnan memmove strchr mktime getrusage gettimeofday getpwnam getgrnam)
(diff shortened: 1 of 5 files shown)


oetiker:
looking at the patch set, I wonder, what is hindering us to use malloc graphfile all the time? Since the code is necessary anyway for the hurd support ...

nirgal:
Considering the patch is quite ugly - I don't know about you, but I don't like having #defined `struct image_desc_t` very much - plus the fact that graph is an heavy function - Just imagining how cairo looks for available fonts makes by blood pressure goes up - I don't think saving a malloc is worth the trouble.
I mean, it's one malloc once per file being graph'ed.

I'll add a test regarding the printf("%s", NULL).

oetiker:
so change the patch and make graphfile dynamic in all cases ... nothing lost ... just maloc MAXPATH if it is available. Then the ifdefs are much fewer

nirgal:
Here you are.

I took the liberty to use strdup in rrd_graph, since it already was used bellow.

Also, I noticed that graph_desc_t.rrd contains a hard-coded length of 1024. This is not an issue for hurd, but this probably is sub-optimal.

oetiker:
thanks

PR #645, opened 2015-08-03
new units=iec and binary options

jrchamp:
Here's the progress that I made trying to adding units=iec; see: https://en.wikipedia.org/wiki/Binary_prefix

You likely have an opinion on how it should be implemented and how the options should interact.

Please take this and run with it, because I won't be able to get back to this in the near future. Thank you!

### DIFF
--- a/bindings/python/rrdtoolmodule.c
+++ b/bindings/python/rrdtoolmodule.c
@@ -274,6 +274,7 @@ static char PyRRD_graph__doc__[] =
     "[-f|--imginfo formatstring] [-a|--imgformat GIF|PNG|GD] "
     "[-B|--background value] [-O|--overlay value] "
     "[-U|--unit value] [-z|--lazy] [-o|--logarithmic] "
+    "[-K|--binary] "
     "[-u|--upper-limit value] [-l|--lower-limit value] "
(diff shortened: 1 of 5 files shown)


oetiker:
hmmm ... I would have expected to see --base=1024 since this should take care of all that is required except for the extra 'i'.

jrchamp:
Yeah, I probably got carried away trying to mirror the behavior of the SI prefix code.

oetiker:
I would rather like to see less new code ... or is that anything your code will accomplish that `--base=1024` and a new switch to change the `?` to `?i`  would not?

oetiker:
can you make an update ?

oetiker:
will you make an update, or should I close this for now ?

jrchamp:
Sorry, I'm not able to work on this now. If your desire is to close this issue, feel free.

### Notes for future work

There are some differences between IEC and SI:
* The power is 1024 instead of 1000.
* The K in IEC Kibi is always capitalized. The k in SI kilo is never capitalized.
* There are no prefixes to represent orders of magnitude smaller than a whole bit or byte and so must not be represented as a fractional portion of the base unit.

oetiker:
shelving this for now

PR #646, opened 2015-08-04
Don't use PATH_MAX in rrd_client.c

nirgal:
This avoids the use of PATH_MAX in rrd_client.c

get_path() now only takes one parameter, and the result must be free()'d.

I decided to free the result as soon as possible in rrdc_update, rrdc_filebased_command, rrdc_info, rrdc_last, rrdc_first, rrdc_create_r2, and rrdc_fetch since it avoid adding many free in all the error handling. Maybe gotos to a bailout would be nicer?

The relative cost of that change is heavier, since in most cases we don't graph. There already were a bunch of strrchr, strdup and so on, but it most cases that does add a malloc/free, and the socket handling must bring its lot too.

I wonder if a simple #ifndef PATH_MAX #define PATH_MAX 4096 wouldn't do the trick. It sure is a extra limitation on hurd, but that doesn't seem unreasonable, and it's so much simplier.

I tried the patch with valgrind and a bunch of existing/unexisting files/dir, and it seems ok to me.

### DIFF
--- a/src/rrd_client.c
+++ b/src/rrd_client.c
@@ -84,15 +84,17 @@ static size_t inbuf_used = 0;
  * into existing solutions (as requested by Tobi). Else, absolute path names
  * are not allowed, since path name translation is done by the server.
  *
+ * The caller must call free() on the returned value.
+ *
  * One must hold `lock' when calling this function. */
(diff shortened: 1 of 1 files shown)


oetiker:
thanks

PR #649, opened 2015-08-07
Allow compilaton on hurd: Stop using PATH_MAX in rrd_daemon.c

nirgal:
This is a work in progress. Just running the travis tests. Please do no merge yet.

### DIFF
--- a/src/rrd_daemon.c
+++ b/src/rrd_daemon.c
@@ -134,7 +134,7 @@ typedef enum { RESP_ERR = -1, RESP_OK = 0, RESP_OK_BIN = 1 } response_code;
 struct listen_socket_s
 {
   int fd;
-  char addr[PATH_MAX + 1];
+  char *addr;
   int family;
 
   /* state for BATCH processing */
@@ -3119,6 +3119,7 @@ static void free_listen_socket(listen_socket_t *sock) /* {{{ */
 
   free(sock->rbuf);  sock->rbuf = NULL;
   free(sock->wbuf);  sock->wbuf = NULL;
+  free(sock->addr);  sock->addr = NULL;
   free(sock);
 } /* }}} void free_listen_socket */
 
@@ -3344,8 +3345,7 @@ static int open_listen_socket_unix (const listen_socket_t *sock) /* {{{ */
 
   listen_fds[listen_fds_num].fd = fd;
   listen_fds[listen_fds_num].family = PF_UNIX;
-  strncpy(listen_fds[listen_fds_num].addr, path,
-          sizeof (listen_fds[listen_fds_num].addr) - 1);
+  listen_fds[listen_fds_num].addr = strdup(path);
   listen_fds_num++;
 
   return (0);
@@ -3578,9 +3578,7 @@ static void open_listen_sockets_traditional(void) /* {{{ */
   }
   else
   {
-    strncpy(default_socket.addr, RRDCACHED_DEFAULT_ADDRESS,
-        sizeof(default_socket.addr) - 1);
-    default_socket.addr[sizeof(default_socket.addr) - 1] = '\0';
+    default_socket.addr = strdup(RRDCACHED_DEFAULT_ADDRESS);
 
     if (default_socket.permissions == 0)
       socket_permission_set_all (&default_socket);
@@ -3599,6 +3597,7 @@ static int close_listen_sockets (void) /* {{{ */
 
     if (listen_fds[i].family == PF_UNIX)
       unlink(listen_fds[i].addr);
+    free(listen_fds[i].addr);
   }
 
   free (listen_fds);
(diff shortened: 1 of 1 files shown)


src/rrd_daemon.c
```diff
@@ -1195,43 +1195,38 @@ static int check_file_access (const char *file, listen_socket_t *sock) /* {{{ */
...
   /* file must be of the format base + "/" + <1+ char filename> */
-  if (strlen(file) < _config_base_dir_len + 2) goto err;
-  if (strncmp(file, config_base_dir, _config_base_dir_len) != 0) goto err;
-  if (*(file + _config_base_dir_len) != '/') goto err;
+  if (strlen(file) < _config_base_dir_len + 2) return 0;
+  if (strncmp(file, config_base_dir, _config_base_dir_len) != 0) return 0;
+  if (*(file + _config_base_dir_len) != '/') return 0;
 
   return 1;
-
```
oetiker: why are you dropping this ?

oetiker:
the code looks fine from what I can see ... the only thing I wondered about ... now that MAX_PATH is not in the picture anymore, could it be that someone causes a problem by supplying an overly long path name to the daemon which it then tries to pass into filesytem functions that are not immune against overly long strings being passed in a path names ?

nirgal:
Tobias Oetiker wrote:
> the code looks fine from what I can see ... the only thing I wondered about ... now that MAX_PATH is not in the picture anymore, could it be that someone causes a problem by supplying an overly long path name to the daemon which it then tries to pass into filesytem functions that are not immune against overly long strings being passed in a path names ?

I did not finish yet, and I don't think I'll do it today.

Well, on the systems that don't have a PATH_MAX, there should not be any
problem.

On the systems that do have a limitation - Linux kernel has a 4096 bytes
hard limit I think, but there may be additional limitations due to file
system, like 8.3 format from iso9660 - this is supposed to fail.

I just tried fopen()'ing for writing a 5k long file: The open fails and
errno is set to 96 (ENAMETOOLONG). perror(3) prints "File name too
long". The error is the same for read and write opening.

My guess is that libc is armored against all this.
ENAMETOOLONG is posix 1 compliant.

I'll have to rework last commit. If I remember correctly some weird
compilers like microsoft ones don't support this:
size_t len = 10;
char path[len];

oetiker:
ok ... will wait with merging then ... the part about MAX_PATH makes sense

nirgal:
Tobias Oetiker:
> ok ... will wait with merging then ... the part about MAX_PATH makes sense

Meaning some other path are obscure I suppose?

I did 2 commits unrelated to PATH_MAX, but related to error handling:

commit:48f416818aedf267b9a64f976ca47306344e41dd is about the
check_file_access:
When there is an error in a handler, like the file doesn't exist or
whatever, a response message is sent with code RESP_ERR, and a string
message.
The handler sends the error and return non-0 if and only if the message
could not be sent.
Such an return code is critical because it will result in the socket
being closed!
Now almost everywhere, there is stuff like `return send_response(sock,
RESP_ERR, "Internal error.\n");`
The one exception is (was) check_file_access: That function does send
the response, so there is no need for the caller to send the response
itself.
There was a problem with handle_request_info, handle_request_first,
handle_request_last, and handle_request_create that were having
duplicates send_response(sock, RESP_ERR, EACESS)
So this commit removed the duplicate and add some comments about the
fact that the send_response from within `check_file_access` is (wrongly)
assumed to have succeeded.

commit:4bd1f75c39f2a5575cfef5d653a7de5c148d757c fixes that.
check_file_access is no longer sending the error response itself, ***nor
ignoring whether that response actually was sent or not*** .
The check_file_access caller is sending the response itself, and thus is
able to see whether it failed or not.
After that commit, all error handling is the same: If send fails,
handler return value is an error (for both regular response message and
for error response message).

Yes, the second commit kind of negates the first. Sorry about that.
This is all more complex than I anticipated, so I tried to have
not-too-bit commits. I hope it makes more sense now.

oetiker:
oh sorry, I just mentioned MAX_PATH because I asked about it and you did respond to that question .. the other parts are fine ... but thanks for the extended explanation, even better now ... its great to see code quality improved.

nirgal:
While cleaning up journal_init, I realised the journal mutex was not released on errors. I fixed that in a separate commit: 31300913220b58864ae545bc3d138babfc82f520
Also, I rewrote last commit, as announced, so that char path[len] is not allocated on the stack, which should help microsoft users (non C99 compliant).

There still is one instance of PATH_MAX left. Please be patient. One more commit to go.

nirgal:
This now compiles on hurd, with tests. \o/
valgrind show some unrelease memory, but so far it looks ok (globals).
This is a big commit. I do hope I didn't break anything.

oetiker:
thanks

PR #650, opened 2015-08-09
Keep CPPFLAGS when building tcl bindings

nirgal:
If CPPFLAGS=-D_FORTIFY_SOURCE=2 during configuration, tcl compilation was not fortified.
This patch adds the missing variable.


oetiker:
thanks

PR #651, opened 2015-08-09
Stop using rpath in binding if it's disabled

nirgal:
I tried moving AM_CONDITIONAL(ENABLE_RPATH) up in the configure.ac, but it has to be quite late.

I did not commit the Makefile.in and other temporary files that depends on that change. You might have to run autoreconf and do it.


oetiker:
thanks

PR #652, opened 2015-08-09
Limit unit tests check to 7 digits precision

nirgal:
Note that 8 digits is not enough a reduction:

on hurd-i386: echo "3.417131125e-02" | perl -p -e 's/([-+]?\d\.\d+e[-+]\d+)/sprintf("%0.8e",$1)/ge'
3.41713112e-02

on armel: echo "3.417131125e-02" | perl -p -e 's/([-+]?\d\.\d+e[-+]\d+)/sprintf("%0.8e",$1)/ge'
3.41713113e-02


oetiker:
not happy ?

nirgal:
I uploaded in Debian a version that doesn't build (fails the test), and I don't like that. :/

I just added this commit as a patch and uploaded as version 1.5.4-2.  I only tested amd64, hurd-i386, and armel. Tests on different architectures should show at https://buildd.debian.org/status/package.php?p=rrdtool&suite=unstable momentarily.
Hopefully, that will do it. :)
Maybe you can check there tomorrow if it's ok everywhere, before merging.
Note that rpn2 test is disabled.

Thank you very much for releasing a new version now, it makes things much easier. I'm happy about that! :)

oetiker:
oh rpn2 is all the fun of the new release ...

nirgal:
Ops, I just understood. The deletion of the branch was just a mistake. :)

oetiker:
so did you figure out what the problem with rpn2 is/was ? my guess is that the detection of what the first day of the week is falls flat ...

nirgal:
> my guess is that the detection of what the first day of the week is falls flat ... 

Yes, that was it. :)

Adding `LANG=C.UTF-8 LC_ALL=C.UTF-8 TZ=UTC` to `make check` parameters did the trick.

oetiker:
oops ... right ... fixed in 60aa415

Issue #658, opened 2015-08-11
(minor) Incorrect permission of RRDCached.pm


nirgal:
Hi

After installation, RRDCached.pm has permission 755 (executable).
However, that file is neither a ELF nor starts with #!, so it should not.

I don't know much about perl. So my guess is that:
- If you'd like to make it an executable, it should be renamed to .pl, and a proper shebang should be added.
- If you're happy to keep it as a perl module and not a perl script, then in `examples/rrdcached/Makefile.am`, it should be moved from section `examples_SCRIPTS` to `examples_DATA`

oetiker:
yes ... it does not have to be executable ... do you want to patch, or should I (will take a bit since I am on another task atm).

nirgal:
I'll do it in a minute

PR #659, opened 2015-08-11
Don't install RRDCached.pm as an executable

nirgal:
This fixes issue #658


nirgal:
I renegerated the Makefile.in
I'm not sure we have the same autoconf version. Make sure you run `autoreconf -fi` and overwrite it if needed.

oetiker:
:-) will do ...

PR #660, opened 2015-08-12
Lua5.2

nirgal:
This add support for lua5.2

The first patch is from Ubuntu. It migrates from lua5.1 to 5.2.

The second patch adds some conditionals so that lua 5.0 and 5.1 still work, along with 5.2.

Reminder for testers: In 5.2, you need `rrd=require("rdd")` while before `require("rrd")` was enough.


oetiker:
thanks!

PR #662, opened 2015-08-13
Support 390x format of nl_langinfo(FIRST_WEEKDAY)

nirgal:
nl_langinfo returns non-portable format for FIRST_WEEKDAY. See
https://www.gnu.org/software/libc/manual/html_node/The-Elegant-and-Fast-Way.html#The-Elegant-and-Fast-Way

That patch adds support for the format used on 390x architecture.
This permits the unit tests to work there.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -526,6 +526,8 @@ AC_LANG_POP(C)
 dnl is time_t 32 of 64 bit ?
 AC_CHECK_SIZEOF([time_t])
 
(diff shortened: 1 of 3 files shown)


nirgal:
I added a test in configure.ac. Maybe there is a better way?
If we just test value>>32, we have warnings on arm.

oetiker:
way to go!

PR #663, opened 2015-08-15
Fixed perl-shared default path

nirgal:
This is only used if you use makemaker directly in the perl-shared directory.
Usually, bindings/Makefile will set up the right value.


oetiker:
thanks

PR #664, opened 2015-08-19
Allow NULL socket addr, as created by systemd

nirgal:
This fixes an error when rrdcached is started using systemd.

Thanks to Denis Bychkov for the diagnostic and the first version of the patch.


oetiker:
thanks

PR #665, opened 2015-08-26
Fixed compilation when there's no sigwaitinfo

nirgal:
* Fixed pointer in call to sigwait
* Don't print pid when we don't have it
* Fixed unused variable warning

### DIFF
--- a/src/rrd_daemon.c
+++ b/src/rrd_daemon.c
@@ -335,16 +335,16 @@ static void sig_common (const char *sig) /* {{{ */
 
 static void* signal_receiver (void UNUSED(*args))
 {
-  siginfo_t signal_info;
   int status;
 
   while (1)
   {
 #if defined(HAVE_SIGWAITINFO)
+    siginfo_t signal_info;
     status = sigwaitinfo(&signal_set, &signal_info);
 #elif defined(HAVE_SIGWAIT)
     status = -1;
-    if (sigwait(signal_set, &status) < 0 ){
+    if (sigwait(&signal_set, &status) < 0 ){
        status = -1;
     }
 #else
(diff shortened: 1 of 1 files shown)


nirgal:
The patch above is ok. Its usage should be triggered on hurd, but it's not.
sigwaitinfo is detected there, while it should not.
This is because of another issue in `configure.ac`:
```
AC_CHECK_FUNCS(sigwaitinfo, [],  AC_CHECK_LIB(rt, sigwaitinfo, [LIBS="${LIBS} -lrt"; AC_DEFINE(HAVE_SIGWAITINFO)],[]))
```

The problem is that hurd is defining a stub, and that sigwaitinfo is defined in librt. But it only returns "Not implemented".

I talked to a hurd specialist and here is his answer:
```
Hello,

As I expected, sigwaitinfo is not implemented, and the symbol you get is
the glibc stub.  This is howver announced in
/usr/include/i386-gnu/gnu/stubs.h:

#define __stub_sigwaitinfo

and detected by autoconf, see config.log:

| #if defined __stub_sigwaitinfo || defined __stub___sigwaitinfo
| choke me
| #endif

but that's only when using AC_CHECK_FUNCS.  When using AC_CHECK_LIB, it
doesn't do this kind of check.  In a way, that's normal: AC_CHECK_LIB
is made to check for the presence of a lib, not for the presence of a
symbol (and actually sigwaitinfo is not even in librt...).  Ideally
the AC_CHECK_FUNCS macro should still be used, with a modified LIBS.
```

I'm quite bad at all this configure stuff. Does any one has a suggestion about how to fix that?
It seems the inner AC_CHECK_LIB should be replaced by a second call to AC_CHECK_FUNC ?

oetiker:
merging this .. just make an other patch if you find a way to properly identify sigwaitinfo on the hurd.

PR #667, opened 2015-08-27
Unlock the global thread lock while processing rrd

nirgal:
This allows other threads to run while doing I/O on rrds, improving parallelism
on multi-threaded python applications.
Thanks to Enrico Zini for reporting the issue.

### DIFF
--- a/bindings/python/rrdtoolmodule.c
+++ b/bindings/python/rrdtoolmodule.c
@@ -139,12 +139,16 @@ static PyObject *PyRRD_create(
 {
     PyObject *r;
     char    **argv;
-    int       argc;
+    int       argc, status;
 
     if (create_args("create", args, &argc, &argv) < 0)
         return NULL;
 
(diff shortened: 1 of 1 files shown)


oetiker:
the problem is that the getopt library is not threadsave and most rrdtool functions except rrd_xport still use getopt ... so in order to make it all thread save the other functions would also have to be switched over to the new getopt used in rrd_xport ... then the beast would be threadsafe.

oetiker:
for now only rrd_xport is thread save ... if you update your patch to reflect that, OR if you migrate other calls away from getops (see rrd_xport) I'll be glad to merge.

nirgal:
The proper thing to do is to use `optparse` everywhere I suppose.
I've been playing in a [optparse branch](https://github.com/nirgal/rrdtool-1.x/commits/optparse). This is not suitable for merge yet.
Only the fix in optparse.c is good to go, as far as I can tell.

PR #668, opened 2015-08-28
Make tests work on lua 5.1 to 5.3

nirgal:


nirgal:
This fixes the lua-rrd example for lua versions 5.2 and 5.3. It also works with lua 5.1 (but not with lua 5.0).

The old version of the example was working on lua 5.0 and 5.1 only.

The old version is kept using an old name. But I believe it should be removed, together with lua 5.0 support. See https://lists.oetiker.ch/pipermail/rrd-developers/2015-August/004473.html

oetiker:
thanks

PR #677, opened 2015-10-18
Don't crash on invalid variable names in RPN

yath:
When rpn_parse() finds [^\0,] after parsing a token, it returns NULL
without setting an error. This causes rrd_test_error() to return false
and subsequent code will dereference NULL (cf. rrdtool xport
CDEF:foo=foo-bar).

This commit changes the OP_VARIABLE branch in rpn_parse so that in order
to be a variable name, sscanf needs to match the full name, causing a
more meaningful "ERROR: don't understand 'illegal-variable-name'" error
message in that case. Also, I made the return NULL branch set an error
message so rrd_test_error() will succeed.

### DIFF
--- a/src/rrd_rpncalc.c
+++ b/src/rrd_rpncalc.c
@@ -433,6 +433,7 @@ rpnp_t   *rpn_parse(
 
 #undef match_op
             else if ((sscanf(expr, DEF_NAM_FMT "%n", vname, &pos) == 1)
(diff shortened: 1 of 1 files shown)


oetiker:
it seems your patches break the tests

yath:
Oops, sorry. I’ve amended the commit.

oetiker:
thanks

PR #678, opened 2015-10-23
rrd_parsetime now unlocks its mutex also after parsing errors.

ufo-dd:
Hello,
I have found a problematic behaviour in rrd_parsetime. When rrd_xport is called multiple times in a process and there are errors in parsing --start or --end parameters, the process blocks endless because the parsetime_mutex is locked.
The reason is, that try() and panic() leaving the function without unlocking the mutex.

### DIFF
--- a/src/rrd_parsetime.c
+++ b/src/rrd_parsetime.c
@@ -843,8 +843,12 @@ char     *rrd_parsetime(
     /* yes this code is non re-entrant ... so lets make sure we do not run
        in twice */
     mutex_lock(&parsetime_mutex);
+
     char *result = rrd_parsetime_nomt(tspec, ptv);
+
+    /* ok done ... drop the mutex lock */
     mutex_unlock(&parsetime_mutex);
+
     return result;
 }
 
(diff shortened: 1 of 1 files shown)


src/rrd_parsetime.c
```diff
@@ -832,16 +832,28 @@ static char *day(
...
     rrd_time_value_t * ptv)
 {
-    time_t    now = time(NULL);
-    int       hr = 0;
-
     /* yes this code is non re-entrant ... so lets make sure we do not run
        in twice */
     mutex_lock(&parsetime_mutex);
+    char *result = rrd_parsetime_nomt(tspec, ptv);
+    mutex_unlock(&parsetime_mutex);
```
oetiker: shouldn't you then get rid of the original  unlock line ? (1009 after this patch)

ufo-dd: Oh yes, I overlooked this.
Sorry this is my first pull request. How is the workflow now?
And I see, that I also have to think about the EnsureFreeMem at the end.

2015-10-23 15:29 GMT+02:00 Tobias Oetiker <notifications@github.com>:

oetiker:
in that case, congrats  to your first pull! ... as soon as I am happy with the patch I merge and it will appear in 1.5.5.

src/rrd_parsetime.c
oetiker: just update your branch, that will trigger a rebuild

oetiker:
thanks!

PR #679, opened 2015-10-31
Optparse

nirgal:


nirgal:
This is a rather big pull. I hope I didn't miss anything.

There is some work left regarding symbols:

I droped @RRD_GETOPT_LONG@ from librrd.sym.in.in. I don't quite understand all the sym.in.in stuff. This may be ok since src/Makefile.am uses grep -v LIBC_HAS_GETOPT_LONG librrd.sym.in.
But then, what is the point of having it in the first place??
I must be missing something...
Either we need a soname bump for removing theses symbols (maybe on some exotic architecture that didn't have getopt?)
And/Or we need a big clean up in all the sym.in handling.

src/rrd_dump.c
```diff
@@ -528,29 +536,13 @@ int rrd_dump(
 
     /* init rrd clean */
 
-    optind = 0;
-    opterr = 0;         /* initialize getopt */
-
-    while (42) {/* ha ha */
```
nirgal: All the while(42) are gone, I hope Earth won't be detroyed now ;)

nirgal:
Ouch, travis CI was failing with errors like
```
$ /opt/rrdtool-master/share/rrdtool/examples/4charts.pl
ERROR: invalid option -- 'interlace'
```
See https://travis-ci.org/oetiker/rrdtool-1.x/builds/88546826

I fixed all the examples in a 3rd commit.

Typos in rrd graph options were silently discarded. Now they throw an error. It's better from my point of view, but people may be surprised :/

oetiker:
so we can no get rid of librrd_th, or rather make it a symlink to librrd ... always linking against rrd_thread_safe.c ... what do you think ? also the thread locking in the python bindings can be relaxed.

src/rrd_dump.c
oetiker: only time will tell :)

nirgal:
I split the previous big commit:

1da51d8 (optparse_long missing support for lonely '--') and cf0df44 (typo in examples about the --interlaced option) can be merged any time you fell.

The big remaining part about replacing getopt by optparse:
- This branch only contains the replacement in the code (I also changed a few argv into options.argv that I had forgotten).
- I'm about to fill a second pull request about actually removing getopt from the library. These functions were exported in some cases (Solaris if I understand correctly), so that their removal may require a bump.

I guess this branch and this pull request are more suitable for a merge in 1.5. But this still is a big change you might not want in the 1.5 branch...

What needs to be double checked here is the src/Makefile.am changes: I had to add the full librrdupd.la into rrdchi and rrdcached.

Maybe it would make sense to have the optparse_init and optparse_long symbol exported from rrdtool library? Either now or in the other pull request (replacing getopt export).

Note that this merge still breaks old examples (fixed in the first commit) and typos in rrdgraph options no longer are silently discarded.

If this is merged, I will refresh the python pull request. Also, some people have been requiring python3 support, I started to work on that a few months back.
Please tell me your plans.

I propose we discuss the librrd_th killing in the other pull request.

oetiker:
ok here we go ...

PR #681, opened 2015-11-02
Nogetopt

nirgal:
That branch focus on dropping getopt support. This can only be considered after #679.

configure.ac was cleaned a bit. Conditional compilation on the embeded version in librrd is droped.

This may require a soname bump for solaris. Please advise... I don't quite know how to do it properly.

As a side effect, the .sym file no longer is dynamic. I adjusted the configure system for that.


nirgal:
Hi

Regarding #683, you wrote
> so this one is the whopper ? containing everything?
> I think I will go to 1.6 when we have this all in, bumping
> the so revision in the process

Not quite. Sorry for the spam. :/

I arranged my pull requests that way:

- The base optparse request is #679. It mostly changes the calls to getopt by calls to optparse. Please double check last note about src/Makefile.am there. That request #679 does not remove getopt on Solaris. It does not require a version bump, as far as I can tell. getopt is still there after #679, it's just unused.

- Two minor pull requests depends on previous one: #682 and #683.

- Python GIL release #667 can also be merged after #679 (branch optparse).

- The real killer is this very PR, #681, aka nogetopt. That one now removes getopt source for the code source.


On the other hand, I would totally understand if you'd rather not merge #679 into 1.5. It's really big.


Now, if we are talking about breaking ABI:
> So we can no get rid of librrd_th, or rather make it a symlink to librrd ... always linking against rrd_thread_safe.c ... what do you think ?

I had not look that far ahead, but yes, you look right. It doesn't make much sense to keep a non thread safe version. The cost is only when calling rrd_strerror(), which should not occur very often; And this is two locks of the thread mutex - one for getting the context, and the other invoking global strerror. In my humble opinion, the benefit of having a more simple code outweigh the very small CPU cost, but I'm sure some people would strongly disagree. Your choice, obviously.

If we are talking about ABI changes, you probably noticed the warning about deprecated dbi API. It looks like the new dbi interface requires a global to be stored in the caller context, so now is the time to make the switch, since rrd_context_t was put inside rrd.h and not inside rrd_tool.h.

Also, on Debian, there are some lintian warnings about 64 bits file offset support being disabled. I have not looked into that one properly, but *if* an ABI change is required, now might be the time to do it too.

Finally, we might want to had a bunch of const before some char* in the ABI. The library will not change the content the string pointed at, as far as I can tell, so this looks ok. I'm talking about compilation warning ones.

I'll try to be quiet today, I promise (that I'll try ;)) Cheers

oetiker:
hihi no problem ... keep your changes coming ... and yes if we break the abi, lets do it right.

oetiker:
it seems #681 contains #679 as well, right ?

nirgal:
Yes it does.
Maybe the situation is more clear if you scroll on the right of the [github network graph](https://github.com/oetiker/rrdtool-1.x/network).
If you are brave, only this #681 "nogetopt" shouldn't be merged in 1.5.

PR #683, opened 2015-11-02
Fix --skip-past-updates support in updatev

nirgal:
As far as I can tell, this was just a missing block in option parsing.

Note that this pull request requires -- I mean contain -- #679.


oetiker:
so this one is the whopper ? containing everything? I think I will go to 1.6 when we have this all in, bumping the so revision in the process

PR #685, opened 2015-11-06
Fix JSON and JSONTIME export with gprints or rules

odoucet:
when you have gprints or rules, json is not formated correctly : a comma is inserted after each block, making two mistakes :
- there is no comma between "legend" and "gprints" block
- there is a trailing comma in "meta" block

So, JSON output is incorrect and does not pass basic checks.

I decided to rewrite this part and add the comma before each block, so this works in all cases : when you don't have any block, only one of them or both.

To reproduce the problem : 
```
/rrdtool graph output.json --imgformat JSONTIME DEF:user=file.rrd:user:AVERAGE XPORT:user#ff0000
/rrdtool graph output.json --imgformat JSONTIME DEF:user=file.rrd:user:AVERAGE XPORT:user#ff0000 VDEF:totalmax=user,AVERAGE GPRINT:totalmax:"%6.2lf"
```

### DIFF
--- a/src/rrd_rpncalc.c
+++ b/src/rrd_rpncalc.c
@@ -415,7 +415,7 @@ rpnp_t   *rpn_parse(
         }
 
         else {
-            rrd_set_error("don't undestand '%s'",expr);
+            rrd_set_error("don't understand '%s'",expr);
             free(rpnp);
             return NULL;
         }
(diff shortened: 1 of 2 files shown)


oetiker:
hi odoucet
could you please make your pull request relative to 1.5 ?

oetiker:
strike that ... I have since brought master forward ... merging now

PR #690, opened 2015-12-02
Add configure option to disable documentation

gustavoz:
Useful for embedded systems where documentation is of no use, or hosts that don't have nroff available.

Signed-off-by: Gustavo Zacarias <gustavo@zacarias.com.ar>


oetiker:
thanks

PR #693, opened 2015-12-20
Use memmove instead of memcpy in rrd_write() to fix undefined behaviour

ntyni:
At least rrdtune ends up calling rrd_write() with the same memory
area for the source and the destination, causing undefined behaviour that has been observed to actually break on the mips architecture.

Presumably the mmap'ed disk IO overhead makes the memmove performance penalty negligible.

Bug-Debian: https://bugs.debian.org/805391
Bug: https://github.com/oetiker/rrdtool-1.x/issues/688


oetiker:
thanks!

PR #699, opened 2016-01-20
new feature for getting a graph's data (instead of the image)

az143:
dear tobi, thanks for rrdtool! here at opmantek.com we really like it and it is serving our (open-source and commercial) needs very well.

however, for one of our commercial products we needed to combine the greatness of rrdtool's data handling with a somewhat more interactive graph display mechanism (ie. highcharts). 

so we've added a function to the shared lib perl interface, RRDs::fetch_graph_pdata, which
does pretty much everything that RRDs::graph does - except, at the end, you don't receive an image but the data _underlying_ the image: the pixel y coordinates, colors, the legends etc.

why did we not just use fetch or xport? the answer is that we need things like CVARs, and we do like the axis autoscaling and all the other graph goodness in rrdtool - we just don't want pngs or gifs. 
fetch_graph_pdata provides that whereas with raw fetch or xport we'd have had to duplicate all the code for scaling, legend, GPRINT etc.

as you and your contributors have been adding improved functionality for external accesses and exportability of rrdtool data in recent releases,  we hope that our fetch_graph_pdata extension fits your plans and makes it into the  official rrdtool code base.

regards
az


oetiker:
neat :) I don't particularly like the interface though ... how about just adding a new graphics format to rrdtool graph the format which would manifest in additional output being generated through the graphv interface ... having another command in the api does not sit all that well with me ...

az143:
agreed; doing this via a new 'raw' graphics format sounds like a better idea. we're not 'married' to this interface at all - this was a relatively quick addon by one of my colleagues when we needed /something/ and /soon/...
i'm not sure how soon we'll find any time to (re)do this more along your preferences; maybe take this to email to discuss more precisely what you'd want the extension to be like?

cheers,
az

oetiker:
you are already using the info protocol to transfer the data if I got this correctly ... so the changes would be pretty straight forward I guess ... I have not studied your code in any detail yet ... 

I am not looking for any standartised graph description format, just less api pollution ...

az143:
tobi, i'm closing this pull request for now - we've realized (finally and somewhat late) that in 1.5.x  graphv with "-a JSON" does about 90% of what our custom code does (as that was originally developed for 1.4.7), and i'll have a look at adding the remaining 10% as optional information to the JSON output.

oetiker:
ok :)

Issue #702, opened 2016-03-23
"make check" fails when VERBOSE is on


nirgal:
The problem comes from the tests/function file. On line 145, we get RRDTOOL_V= then RRDTOOL=verbose_rrdtool. Only after that, RRDTOOL is set to $TOP_BUILDDIR/src/rrdtool. Here's a run with a set -x:
```
++ TZ=CET
++ export TZ
++ BASEDIR=/<<PKGBUILDDIR>>/tests
+++ readlink -f -- /<<PKGBUILDDIR>>/tests
++ BASEDIR=/<<PKGBUILDDIR>>/tests
++ BUILDDIR=/<<PKGBUILDDIR>>/tests
++ TOP_BUILDDIR=/<<PKGBUILDDIR>>
++ DIFF='diff -u'
++ DIFF9=seddif
++ '[' -n 1 ']'
++ RRDTOOL_V=
++ RRDTOOL=verbose_rrdtool
++ '[' -z verbose_rrdtool ']'
+ BASE=/<<PKGBUILDDIR>>/tests/modify-test1
+ BUILD=/<<PKGBUILDDIR>>/tests/modify-test1
+ verbose_rrdtool create /<<PKGBUILDDIR>>/tests/modify-test1a.rrd --start 1300000000 --step 60 DS:a:GAUGE:120:0:U RRA:AVERAGE:0.5:1:10 RRA:AVERAGE:0.5:5:10
+ echo '' create /<<PKGBUILDDIR>>/tests/modify-test1a.rrd --start 1300000000 --step 60 DS:a:GAUGE:120:0:U RRA:AVERAGE:0.5:1:10 RRA:AVERAGE:0.5:5:10
 create /<<PKGBUILDDIR>>/tests/modify-test1a.rrd --start 1300000000 --step 60 DS:a:GAUGE:120:0:U RRA:AVERAGE:0.5:1:10 RRA:AVERAGE:0.5:5:10
+ '' create /<<PKGBUILDDIR>>/tests/modify-test1a.rrd --start 1300000000 --step 60 DS:a:GAUGE:120:0:U RRA:AVERAGE:0.5:1:10 RRA:AVERAGE:0.5:5:10
./functions: line 73: : command not found
+ fail create
+ RC=127
+ echo 'FAILED: (rc=127)' create
FAILED: (rc=127) create
+ '[' -n '' ']'
+ '[' -n '' ']'
+ exit 127
FAIL modify1 (exit status: 127)
```
[...]

PR #703, opened 2016-03-24
Fixed tests when VERBOSE is on

nirgal:
Define RRDTOOL_V from RRDTOOL, but only after RRDTOOL is defined.
Output the header showing witch command is run on stderr, not stdout.

### DIFF
--- a/tests/functions
+++ b/tests/functions
@@ -68,7 +68,7 @@ function valgrind {
 }
 
 function verbose_rrdtool {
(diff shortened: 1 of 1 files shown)


nirgal:
This is a fix for #702 

I'm not sure it's a good idea to have the headers to stderr, but I don"t have a better one right now.

oetiker:
better than break tests. thanks

PR #704, opened 2016-03-26
Update tcl gitignore for version 1.5

nirgal:
I get this warning when I do a `git status`:
```
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	bindings/tcl/tclrrd1.5.999.so
```


nirgal:
Something similar need to be done in master, maybe 1.6 or maybe something more generic...

oetiker:
how about ignoring 

`tclrrd*.so`

oetiker:
3b6f1c7

PR #706, opened 2016-04-10
Disable nagel algorithm

stevenh:
Disable TCP's nagal algorithm which was causing massive slowdown in processing rrdcached requests.

Also check for other setsockopt call errors.


oetiker:
thanks!

PR #709, opened 2016-04-22
Fixed CHANGES about supported lua version

nirgal:
lua 5.1 already was supported.
What version 1.6.0 brings is lua 5.2 and lua 5.3 support.


oetiker:
:-) thanks ... being a luo noob

PR #715, opened 2016-06-21
Output legend coords, for making img map

itensionanders:


oetiker:
hmmm there is the graphv api ...  which would be ideally suited to output extra information ... if you would adapt your patch to work with graphv, I would be glad to merge it

itensionanders:
Ah, right. Maybe it will need some keyword (like "legend"), to stand out from the other information ?

Or how you can best convert it to a `key = value` format, when the keys can be pretty much anything.

graphv
```
graph_left = 67
graph_top = 33
graph_width = 650
graph_height = 300
image_width = 747
image_height = 415
graph_start = 1466268240
graph_end = 1466269260
value_min = 0,0000000000e+00
value_max = 4,0000000000e+00
```

coords
```
  1-min	16,352,65,366
Now: 48,0m	81,352,151,366
Min: 48,0m	167,352,237,366
Avg:236,4m	253,352,323,366
Max:532,0m	339,352,409,366
  CPUs 	16,366,65,380
Now:  4,0 	81,366,151,380
Min:  4,0 	167,366,237,380
Avg:  4,0 	253,366,323,380
Max:  4,0 	339,366,409,380
  Procs	16,380,65,394
Now:  0,0 	81,380,151,394
Min:  0,0 	167,380,237,394
Avg:355,6m	253,380,323,394
Max:  3,1 	339,380,409,394
  test	16,394,58,408
```

itensionanders:
graph

oetiker:
yes convert it to key value and add as many as you need ...maybe: 

```
label_0 = 81,380,151,394,"Now:  0,0"
label_1 = ...
```

oetiker:
OR just have labes = {big json blob}

itensionanders:
Is it "label" or "legend" ? Oh well, easy to change afterwards. Will change to use graphv output first.

oetiker:
legend makes more sense

itensionanders:
Hmm, having to write and then parse JSON sounds like it could get complicated for key=value ?
Found out that you have to type the output, so it will probably be string to look like rect coords:

https://www.w3.org/TR/html-markup/area.html#area.attrs.coords.rect
`  <area shape="rect" coords="16,394,58,408" href="test.html" alt="test">`

itensionanders:
I thought the format used by print was pretty nice, so went with that:
```
legend["  1-min"] = "16,352,65,366"
legend["Now: -nan "] = "81,352,151,366"
legend["Min: -nan "] = "167,352,237,366"
legend["Avg: -nan "] = "253,352,323,366"
legend["Max: -nan "] = "339,352,409,366"
legend["  CPUs "] = "16,366,65,380"
legend["Now: -nan "] = "81,366,151,380"
legend["Min: -nan "] = "167,366,237,380"
legend["Avg: -nan "] = "253,366,323,380"
legend["Max: -nan "] = "339,366,409,380"
legend["  Procs"] = "16,380,65,394"
legend["Now: -nan "] = "81,380,151,394"
legend["Min: -nan "] = "167,380,237,394"
legend["Avg: -nan "] = "253,380,323,394"
legend["Max: -nan "] = "339,380,409,394"
```

oetiker:
the problem with that format is that scripting interfaces are implementing graphv returns as hashes/maps ... if you output data like this, there will be duplicate keys

itensionanders:
Right. Basically I am not interested in those keys (now/min/avg/max) anyway - but it's a problem if it breaks the bindings, like if multiple keys in a map is not allowed. Will do a counter then, but it still remains to find an output format that would work. I want to keep the coords (in usemap format), and need the text in order to find the original key (to find the URL to use)

itensionanders:
Split it up into two. Maybe it is too chatty now, and needs some kind of opt-in ?

```
legend[0] = "  1-min"
coords[0] = "16,352,65,366"
legend[1] = "Now: -nan "
coords[1] = "81,352,151,366"
legend[2] = "Min: -nan "
coords[2] = "167,352,237,366"
legend[3] = "Avg: -nan "
coords[3] = "253,352,323,366"
legend[4] = "Max: -nan "
coords[4] = "339,352,409,366"
legend[5] = "  CPUs "
coords[5] = "16,366,65,380"
legend[6] = "Now: -nan "
coords[6] = "81,366,151,380"
legend[7] = "Min: -nan "
coords[7] = "167,366,237,380"
legend[8] = "Avg: -nan "
coords[8] = "253,366,323,380"
legend[9] = "Max: -nan "
coords[9] = "339,366,409,380"
legend[10] = "  Procs"
coords[10] = "16,380,65,394"
legend[11] = "Now: -nan "
coords[11] = "81,380,151,394"
legend[12] = "Min: -nan "
coords[12] = "167,380,237,394"
legend[13] = "Avg: -nan "
coords[13] = "253,380,323,394"
legend[14] = "Max: -nan "
coords[14] = "339,380,409,394"

```

oetiker:
that does look much nicer :)

oetiker:
now if you can add a test (see the test sub directory) we are good to go I'd say

itensionanders:
@oetiker : seems like there was no graph/output in the existing test, so had to add that first...
Also it seems that the tests always fail for me (due to locale), unless I run them with LANG=C ?

i.e. it differs between periods and commas, as the decimal sign, between sv_SE.UTF-8 and C

itensionanders:
Hmm, it seems like it could be difficult to make the test line up with all different library versions etc...

itensionanders:
Victim to things like https://bugzilla.gnome.org/show_bug.cgi?id=329547 and font-related issues.

oetiker:
yes, adding LANG=en_US.utf8 might be a good thing in the test runner ...

regarding system differences, you may want to 'filter the coordinates' passing the output through a filter that removes the coords numbers from the output prior to comparison. there is also code in the test suite that limits the precision of numeric output to get consistent output between different systems ...

itensionanders:
@oetiker : test now blanks out all coords, replacing the numbers with '#'
Did it needs some kind of toggle parameter too, or is it good to go now ?

oetiker:
no there is no need for a toggle .... thank you very much for the PR

itensionanders:
Excellent! Now all that is left is to parse the thing back in again (using PHP), and then add the URLs...
I suppose it would be fair to say that using this feature (legend/coords) requires rrdtool 1.6 or later ?

PR #716, opened 2016-06-24
Master coverity

Marek77:
Hello Tobi,

we are using rrdtool at work and as we have Coverity for static code analysis we want to contribute back :-)

Please don't hesitate to contact me if you have doubts with some of the commits, the log messages could be more verbose especially for the less-than obvious fixes.

regards
Marek

### DIFF
--- a/src/rrd_cgi.c
+++ b/src/rrd_cgi.c
@@ -1354,6 +1354,26 @@ char     *rrdcgiDecodeString(
     return text;
 }
 
+/* free_result(s_var**, number)
+ *
+ * Clean-up the 'result' variable from rrdcgiReadVariables() properly.
+ * We can safely call free() on result[i]->{name,value} because they are
+ * memset() to 0 after their allocation.
+ */
+static void free_result(s_var **result, int number)
+{
+    int i;
+
+    for(i = 0; i < number; i++) {
+        if (result && result[i]) {
+            free(result[i]->name);
+            free(result[i]->value);
+            free(result[i]);
+        }
+    }
+    free(result);
+}
+
 /*  rrdcgiReadVariables()
  *
  *  Read from stdin if no string is provided via CGI.  Variables that
@@ -1379,8 +1399,10 @@ s_var   **rrdcgiReadVariables(
             length = atoi(ip);
             if ((line = (char *) malloc(length + 2)) == NULL)
                 return NULL;
-            if (fgets(line, length + 1, stdin) == NULL)
+            if (fgets(line, length + 1, stdin) == NULL) {
+            	free(line);
                 return NULL;
+            }
         } else
             return NULL;
     } else if (cp && !strcmp(cp, "GET")) {
@@ -1405,6 +1427,10 @@ s_var   **rrdcgiReadVariables(
                         return NULL;
                     strncat(line, tmp, tmplen);
                 } else {
+                    /* clean-up the storage allocated in previous iteration */
+                    if (line)
+                        free(line);
+
                     if ((line = strdup(tmp)) == NULL)
                         return NULL;
                 }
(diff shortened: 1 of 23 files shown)


src/rrd_cgi.c
```diff
@@ -418,7 +418,12 @@ static int readfile(
         totalcnt = (ftell(input) + 1) / sizeof(char) - offset;
         if (totalcnt < MEMBLK)
             totalcnt = MEMBLK;  /* sanitize */
-        fseek(input, offset * sizeof(char), SEEK_SET);
+        if (fseek(input, offset * sizeof(char), SEEK_SET) == -1)
```
stevenh: Slight style issue here.

src/rrd_cgi.c
```diff
@@ -1379,8 +1406,10 @@ s_var   **rrdcgiReadVariables(
             length = atoi(ip);
             if ((line = (char *) malloc(length + 2)) == NULL)
                 return NULL;
-            if (fgets(line, length + 1, stdin) == NULL)
+            if (fgets(line, length + 1, stdin) == NULL) {
+            	free(line);
```
stevenh: Indentation (could be github visualisation of it though)

src/rrd_dump.c
```diff
@@ -569,6 +569,7 @@ int rrd_dump(
                           "[--no-header|-n]\n"
                           "[--daemon|-d address]\n"
                           "file.rrd [file.xml]", options.argv[0]);
+            if (opt_daemon) free(opt_daemon);
```
stevenh: Personally no a fan of this style of shorthand but not sure what the project rules are about it. That and it should be a NULL test.
```c
if (opt_daemon != NULL)
   free(opt_daemon);
```

oetiker: *personally* I would even make the curly brackets ... :) as you can see the code is rather 'old' by now ...

oetiker:
cool! thank you very much ... while on the subject ... isn't there some way to integrate coverty on travis and they hand out free licenses to oss projects ... do you have any experience with that ?

src/rrd_cgi.c
Marek77: Please expand?

src/rrd_dump.c
Marek77: I have seen all kinds of code style in the source ;-) but I'm happy to adapt. Let me know..

Marek77:
Sorry no idea how Github + Travis + Coverity works...

src/rrd_dump.c
oetiker: hihi ... yes ... so please ... braces everywhere ...

src/rrd_cgi.c
Marek77: Yeah my editor removes all whitespace so I had to hand-edit some commits not to add a lot of whitespace diff. There should be 4 spaces there.

oetiker:
is there a coverty configuration file ? or something like that ?

src/rrd_xport.c
```diff
@@ -444,15 +444,17 @@ int rrd_graph_xport(image_desc_t *im) {
...
   }
+
+  /* free legend */
+  for (unsigned long j = 0; j < col_cnt; j++) {
+    free(legend_v[j]);
+  }
+  free(legend_v);
+  /* free data */
+  free(data);
+
```
oetiker: are you sure that it is ok todo this in any event (regardles of r?)

Marek77: Actually if (r != 0), free(data) is done and then return; however "data" is allocated by rrd_xport_fn() (line 407), (r==0) (switch(im->imgformat) case default)  but not free'd without this patch.

Marek77:
I don't know about the configuration files; we have a coverity installation and "just use it" :-) (but I need to have a product / rrdtool,... / configured in coverity) but it's the admin team who manages it. Then I send my "make install" via cov-build wrapper.

Marek77:
I have added curly braces ;-) and fixed the indent as per stevenh's comments.

PR #729, opened 2016-08-23
src/rrd_open.c: fix crash (SIGBUS) trying to CREATE a rrd on a full fs

Marek77:
This patch fixes a SIGBUS observed in testing. Jump to cleanup (goto out_close) on error instead of using the invalid file descriptor for lseek etc.


oetiker:
thanks!

PR #730, opened 2016-08-24
m4/acinclude.m4: fix for build error 'possibly undefined macro: AS_VAR_COPY

Marek77:
Fix build on RHEL6 / autoconf 2.63 or older.


oetiker:
thanks!

PR #732, opened 2016-08-24
Fix "mkdir -p"

Marek77:
This PR fixes a race condition when using rrdcached with "-R" option with several threads trying to mkdir the same directory (or simply if the directory already exists), and improves the log message in case of mkdir failure.


oetiker:
thanks

PR #741, opened 2016-09-22
fixed typo for 100,000 second in example1

sharuzzaman:
the example show heartbeat of 100,000 but the explanation show 10,000


oetiker:
thanks!

PR #742, opened 2016-09-24
Improve the RRD client API

tokkee:


oetiker:
could you update librrd.pod too

tokkee:
I updated librrd(3).

While doing that I wondered if the `rrd_client_*_if_daemon` functions actually make sense. I don't actually think so because the `rrd_client_` functions don't make any sense if not connected to a daemon in the first place, hence, they will be used differently.

Also, the `addr` parameter to `rrd_client_is_connected` does not make overly sense. Hence, I suggest dropping it and removing `rrd_client_is_any_connected`. That'll be a little inconsistent with the `rrdc_` interface but I think that's fine and makes more sense that way (the two interfaces are different after all which is why we have them ;-)).

Thoughts?

oetiker:
sure ... the nicer the new set of functions the better :)

oetiker:
I would be ready to merge this ... ok ?

tokkee:
Okay, I think this is good to go now :-) Sorry for the delay.

oetiker:
thanks for the contrib!

PR #744, opened 2016-10-04
Read the last timestamp from the RRD file, instead of setting it to z…

bangert:
…ero.

This fixes a dataloss problem when trying to insert old data. rrdcached
will accept the data at first, but then fail a whole batch when trying
to flush the data to disk.

The last timestamp is initially set to 0, and later updated to the
timestamp of the first update. However, if the RRD file already contains
newer data, the flush will fail later on. And since RRD cache writes all updates
in one go, the whole batch of samples fails and is discarded (even though,
just the first few updates are too old).

By reading in the last timestamp from the RRD file, we already fail
correctly on the individual update commands (ie. on the protocol layer).
```
RRD  ----------------------|
RRDcached         |------------------| <- Flush
                           |---------| <- Dataloss
```

With this patch, the initial updates, that overlap with data in the RRD
file will fail individually (at submission time).
```
RRD  ----------------------|
RRDcached         |xxxxxxxx----------| <- Flush
```

Thanks

### DIFF
--- a/src/rrd_daemon.c
+++ b/src/rrd_daemon.c
@@ -1594,6 +1594,37 @@ static int handle_request_update (HANDLER_PROTO) /* {{{ */
       goto done;
     }
 
+    time_t last_update_from_file;
+    rrd_file_t * rrd_file;
+    rrd_t rrd;
+
+    rrd_clear_error();
+    rrd_init(&rrd);
+    rrd_file = rrd_open(file, &rrd, RRD_READONLY);
+    if (!rrd_file)
+    {
+      rrd_free(&rrd);
+      free (ci);
+      RRDD_LOG (LOG_ERR, "handle_request_update: Could not read RRD file.");
+
+      rc = send_response(sock, RESP_ERR, "RRD Error: %s\n", rrd_get_error());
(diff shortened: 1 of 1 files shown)


oetiker:
have you looked at the performance impact of your change ? looking at your code I get the impression that the daemon will read the rrd file on every update to do it's checks ...

it is important to design the code such that the performance in the non-error case is impacted as little as possible ... it is great to handle errors gracefully, but the normal use case should not suffer a performance penalty for recovering errors that should not happen in the first place.

bangert:
we have not performed any tests. nevertheless, it is not as you suspect an rrd_open on every update. the additional code is in the `if (ci == NULL)` block, so it is only executed if we have no information about the rrd at hand, so only once.

i must confess, i dont know how expensive rrd_open is, but my optimistic guess is that we only need to read one addional block. the rrd file is stat'ed a couple of lines earlier, so most of the fs layer info has already been loaded.

however, one could possibly optimize this section, in areas not touched upon by this pull request: by only stat'ing the rrd file, if there is no `ci` object and not on every update. haven't given this much thought, though...

Issue #749, opened 2016-10-18
Mis-handling timestamps when sizeof(time_t) != sizeof(long) (?)


he32:
It seems that rrdtool 1.60.0 mis-handles timestamps in an rrd database when it does
"rrdtool dump" when the host system has a time_t type which is larger than long, as is
the case on NetBSD/i386 where time_t is 64 bits, while long is 32 bits.

A recently created rrd database filled with a few samples and given to "rrdtool dump" produces

```
                    <!-- 2152-11-24 22:28:16 CET / 1476802800 --> <row><v>1.113549608e+02</v><v>1.069877135e+01</v></row>
                    <!-- 2152-11-24 22:58:16 CET / 1476804600 --> <row><v>1.202089651e+02</v><v>8.446735203e+01</v></row>
                    <!-- 2152-11-24 23:28:16 CET / 1476806400 --> <row><v>1.506548764e+05</v><v>4.057508256e+06</v></row>
                    <!-- 2152-11-24 23:58:16 CET / 1476808200 --> <row><v>2.785666863e+05</v><v>3.966554052e+07</v></row>
                    <!-- 2152-11-25 00:28:16 CET / 1476810000 --> <row><v>1.744465412e+05</v><v>2.087267285e+07</v></row>
                    <!-- 2152-11-25 00:58:16 CET / 1476811800 --> <row><v>7.064252154e+06</v><v>1.907582096e+07</v></row>
                    <!-- 2152-11-25 01:28:16 CET / 1476813600 --> <row><v>3.792049204e+06</v><v>1.311867946e+07</v></row>
                    <!-- 2016-10-18 20:30:00 CEST / 1476815400 --> <row><v>3.071464893e+06</v><v>1.002659453e+07</v></row>
            </database>
```

Only the last timestamp is correctly decoded into something similar to today's date.

No fix supplied as of yet.

(it looks like the graphing from this data isn't affected by this so far apparently cosmetic issue)

he32:
Debugging reveals that this calculation is problematical:

    422                 now = (rrd.live_head->last_up
    423                        - rrd.live_head->last_up
    424                        % (rrd.rra_def[i].pdp_cnt * rrd.stat_head->pdp_step))
    425                     + (timer * rrd.rra_def[i].pdp_cnt * rrd.stat_head->pdp_step);

[...]

PR #750, opened 2016-10-18
Fix a signedness/unsignedness conversion bug in the expression

he32:
calculating "now" before printout of the timestamp.  Negative values
are first cast to unsigned, it seems, causing overflow.

This is the fix belonging to issue #749, where the title is actually
misleading.


oetiker:
thanks

PR #751, opened 2016-10-20
1.6 windows x64

spartawhy117:
modify and add some files to compile v1.6 with vs2015 on x64.

some explain has been add to win32-build-tips.txt. because some modify has big changes that effect linux,so this branch really need a new branch rather than merge.


.gitignore
```diff
@@ -1,72 +1,242 @@
...
-/doc/RRD?.pod
-/contrib
-/win32/Debug
-/win32/Release
-/win32/ReleaseDLL
-/win32/*.suo
-/win32/*.sdf
-/bindings/dotnet/*.suo
-/win32/*.opensdf
+*.pgc
```
oetiker: could you actually merge and not just replace

oetiker:
ping

spartawhy117:
however i found one exception when run exe. that maybe the dll problem of libxml2-2. 
however, i finally compiled 1.4 version on windows-64 with vs2015.

Issue #698, opened 2016-01-14
Consolidate PyPi package naming


macfreek:
This is admittedly more of administrative request than anything else, please bear with me, while I try to make my point.

It seems that are multiple sources providing Python bindings to RRDtool, with often different naming schemes in the packages for different distributions. I'm familiar with the following sources:

Name | Pip package | Version | Year | Project URL | Repository URL | Author
---- | ----------- | ------- | ---- | ----------- | -------------- | -----------
py-rrdtool |  | 0.2.1 | 2003 | http://www.nongnu.org/py-rrdtool/ | [_rrdtoolmodule.c](http://cvs.savannah.gnu.org/viewvc/py-rrdtool/py-rrdtool/src/_rrdtoolmodule.c?view=annotate) | Hye-Shik Chang
py-rrdtool | py-rrdtool | 1.0b1 | 2006 | http://py-rrdtool.sourceforge.net/ | [_rrdtoolmodule.c](http://py-rrdtool.cvs.sourceforge.net/viewvc/py-rrdtool/py-rrdtool/src/_rrdtoolmodule.c?view=annotate) | Hye-Shik Chang, Mihai Ibanescu
 | python-rrdtool | 1.4.7 | 2012 | https://github.com/pbanaszkiewicz/python-rrdtool | [rrd_tool.c](https://github.com/pbanaszkiewicz/python-rrdtool/blob/master/rrdtool-1.4.7/src/rrd_tool.c) | Piotr Banaszkiewicz
rrdtool (previously: rrdtoolmodule) |  | 1.5.5 | 2015 | http://oss.oetiker.ch/rrdtool/prog/rrdpython.en.html | [rrdtoolmodule.c](https://github.com/oetiker/rrdtool-1.x/blob/master/bindings/python/rrdtoolmodule.c) | Hye-Shik Chang, Alan Milligan
rrdtool | rrdtool | 0.1.2 | 2015 | https://github.com/commx/python-rrdtool | [rrdtoolmodule.c](https://github.com/commx/python-rrdtool/blob/master/rrdtoolmodule.c) | Hye-Shik Chang, Christian Jurk

[...]

PR #755, opened 2016-11-07
python-rrdtool as replacement for the upstream Python bindings

commx:
As already mentioned in #698, various Python bindings for rrdtool exist. I've been maintaining [python-rrdtool](https://github.com/commx/python-rrdtool) for some time now and it got appropriate Python 2.x and 3.x support, and I have synchronised code changes with upstream for most things.

I've made some changes to the m4 files to enable Python 3 support and the bindings successfully build for Python 2 and 3 now (make sure to set the desired interpreter version as default, symlinking `/usr/bin/python` to `/usr/bin/python3.5` on Ubuntu 16.04 for example).

Aside from this, I'll continue to maintain python-rrdtool as-is, because it allows users to install it over PyPI which is the preferred way to install Python packages and it makes things much easier especially when using virtualenvs.


oetiker:
cool ... will be glad to merge ... but before I do, I question. Have you looked into supporting the new callback interface ? there is only an implementation in perl yet, but I guess python users might find this cool too ... look at the callback.t file in the perl binding directory

commx:
I've added the code for the callback interface and also added it to python-rrdtool 0.1.7. A small example on how to use it can be found in the documentation for the python-rrdtool docs: http://pythonhosted.org/rrdtool/usage.html#register_fetch_cb

I also added a method called ``clear_fetch_cb`` which can be used to free the callback function and "unregister" the fetch callback.

oetiker:
cool! ... so while you are 'in the flow' how about throwing a test into the mix? so that we will know the bindings keep working whenever changes are made to rrdtool ? see the test directory

commx:
Modified some bits to get the initial tests added.

The minimum required Python version for the bindings shall be 2.7, as Python 2.6 has reached it's End-of-life in October 2013 already. New builds shall always at least Python 2.7 available.

nwf:
Debian is citing this PR for their packaging of the rrdtool python interface (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=780543) and I'd very much like that package to be python3-compatible.  What's the status of this PR?  Could another pair of eyes and hands help?

oetiker:
thanks!

PR #757, opened 2016-12-03
add support for negative logarithms

mmitch:
Define a *signed logarithm* function that works like this:
```
   x >= 0 : slog10(x) =   log10(  x )
   x  < 0 : slog10(x) = - log10( -x )
```
This will allow drawing both above und below the X axis when using logarithmic scale.

Current limitations:
 - Y autoscale does not detect negative values, `--lower-limit` has to be used
 - Y axis labeling is missing

Can anybody help me to solve these problems?

### DIFF
--- a/src/rrd_graph.c
+++ b/src/rrd_graph.c
@@ -242,6 +242,16 @@ const char default_duration_fmt[] = "%H:%02m:%02s";
 #endif
 
 
+/* #define SLOG10(x)
+   "signed" logarithm
+   SLOG(x) = log10(abs(x)) * signum(x) */
+#define SLOG10(x)	((x) < 0 ? -log10(-(x)) : log10((x)))
+
+/* #define SPOW(x)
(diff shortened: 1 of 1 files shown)


mmitch:
see this script for a working demonstration:
https://gist.github.com/mmitch/6a7e4bda057d0b4ae2d46515f5a30f7a

dword1511:
Doesn't slog10(0.1) and slog(-10) get the same value?

mmitch:
Strictly mathematically speaking: yes.
But I have not yet seen an rrdgraph logarithmic diagram with negative values. I think the values are autoscaled in such a way that x never is less than 1 (at least the numeric value of x - the axis legend will happily show a scale of 1/10 or 1/1000 using prefixes like `m` or `u`).

oetiker:
if memory serves I am cutting values <=1 since there is a discontinuity

PR #764, opened 2017-01-25
restart rrdcached on upgrade, start and stop in install/uninstall

tvestelind:
While using this spec file we had problems when upgrading rrdtool-cached. The package had disabled autostarting rrdcached due to the order of which the RPM scriptlets are run (pre, post new package then preun, postun old package) when doing an upgrade.

To mitigate the problem where the old pre- and postun scriplets are executed during and upgrade I added the %posttrans scriptlet which is run absolute last.

Values and recommendation taken from: https://fedoraproject.org/wiki/Packaging:Scriptlets

Signed-off-by: Tomas Vestelind <tomas.vestelind@gmail.com>

### DIFF
--- a/rrdtool.spec
+++ b/rrdtool.spec
@@ -325,6 +325,10 @@ else # Upgrading
     /sbin/service rrdcached restart || :
 fi
 
+%posttrans
+/sbin/chkconfig --add rrdcached || :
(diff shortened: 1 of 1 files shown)


oetiker:
thanks!

PR #771, opened 2017-03-02
use systemd macros for rrdcached in rpm spec file

tvestelind:
I've added:
* a variable to control whether or not to build for systemd
* systemd-provided macros in the RPM spec-file according what Fedora
  recommends here: 
  - https://fedoraproject.org/wiki/Packaging:Scriptlets?rd=Packaging:ScriptletSnippets#Systemd and 
  - https://fedoraproject.org/wiki/Packaging:Systemd#Packaging

Signed-off-by: Tomas Vestelind <tomas.vestelind@gmail.com>

### DIFF
--- a/rrdtool.spec
+++ b/rrdtool.spec
@@ -3,6 +3,7 @@
 %define with_tcl %{?_without_tcl: 0} %{?!_without_tcl: 1}
 %define with_ruby %{?_without_ruby: 0} %{?!_without_ruby: 1}
 %define with_lua %{?_without_lua: 0} %{?!_without_lua: 1}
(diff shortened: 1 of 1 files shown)


oetiker:
thanks!

tvestelind:
No problem :)

I just want to add that I haven't really tried that systemd macro. We define that macro differently in our version of this spec file so if you could look closely at that part it would be great :)

oetiker:
well others willl I am not building RPMs myself

tvestelind:
Ok, good. Thank :)

PR #788, opened 2017-05-29
Added -o option for logging into a file instead of using syslog.

svenpanne:
For some use cases it is better to keep log output separated from the normal
syslog. To handle such scenarios, a new -o option has been added.

Fixed foreground logging on the way, too: There was a missing mutex for
avoiding intermingled output.


oetiker:
thanks!

PR #789, opened 2017-05-29
Do not initialize Cairo/Pango on export.

svenpanne:
On a heavily loaded rrdcached, initializing lots of unnecessary Cairo/Pango
data structures (surfaces, font-related structures, etc.) is not a good
idea. Do this only when we really want to draw something.

### DIFF
--- a/src/rrd_graph.c
+++ b/src/rrd_graph.c
@@ -437,6 +437,7 @@ int im_free(
     }
     free(im->gdes);
 
+    if (im->init_mode == IMAGE_INIT_CAIRO) {
     for (i = 0; i < DIM(text_prop);i++){
         pango_font_description_free(im->text_prop[i].font_desc);
         im->text_prop[i].font_desc = NULL;
@@ -467,6 +468,7 @@ int im_free(
         g_object_unref(im->layout);
     }
     mutex_unlock(im->fontmap_mutex);
+    }
 
 	if (im->ylegend)
 		free(im->ylegend);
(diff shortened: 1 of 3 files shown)


svenpanne:
FYI: The actual change without whitespace changes is quite small, just add ``?w=1`` to the GitHub URL...

oetiker:
could you please push again without whitespace change and to the reformatting in a separate PR if you

svenpanne:
OK, I've changed this PR to include only the non-whitespace changes. The change could even be smaller:

* One can get away without using a new ``enum`` by using an ``int``, but that is a bad idea in general because it makes the call sites quite obscure (What does a ``0`` mean? What is ``1``?). Boolean arguments are widely considered an antipattern already, so let's not make it worse... ;-)

* If I see it correctly, one could avoid a new field in ``image_desc_t`` by checking e.g. its ``font_options`` member against ``NULL`` instead, but this is quite implicit. Making intentions explicit is always better, and the structure is already huge, so size doesn't really matter here.

But if you would like a minimal change, I'll be happy to upload a new version of this PR.

bangert:
Nice. Whats the performance improvement (in rough numbers) you get from this change?

svenpanne:
I don't have any concrete numbers at hand, but we ship a patched rrdcached with our software and use it heavily via our graphical frontend (https://mathias-kettner.de/cms_graphing.html). For larger installations at our customers we know that writing/retrieving RRD data is a kind of bottleneck, seeing substantial I/O and CPU load from the rrdcached. So every little improvement helps us.

In addition, this PR has only a tiny impact compared to an improvement in rrdcached's socket buffering scheme, which has quadratic(!) complexity at the moment. PR upcoming, stay tuned...

Furthermore, if I read our own Git logs correctly, the initial patch upon which this PR is based is from Tobi himself, somewhere in 2015. :-)

svenpanne:
Just out of curiosity: Is something still missing to get this PR merged? Having this patch in upstream would be great, because it would bring our local patches down to zero.

oetiker:
nothing is missing ... I just kept getting distracted while reviewing ... :)

PR #790, opened 2017-05-30
Fixed declaration and export of rrd_dump_opt_r.

svenpanne:
The previous state of affairs regarding rrd_dump_opt_r was a bit
inconsistent: The dynamic library didn't export it, while the static one
did.  So the right way to fix it will be exporting it in both variants,
because removing it from the static library would require a major version
change.  Furthermore, rrd_dump_opt_r is mentioned in the documentation at
librrd.pod, so this is another hint for the fix.

In addition, a C prototype was wmissing, so GCC complained about it.

### DIFF
--- a/src/librrd.sym
+++ b/src/librrd.sym
@@ -11,6 +11,7 @@ rrd_dontneed
 rrd_dump
 rrd_dump_r
 rrd_dump_cb_r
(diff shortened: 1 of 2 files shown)


oetiker:
sigh ...  ok

PR #791, opened 2017-05-30
Do not specify -version-info for convenience libraries.

svenpanne:
libtool warns about a useless -version-info flag for convenience libraries,
see:

   http://git.savannah.gnu.org/cgit/libtool.git/tree/build-aux/ltmain.in#n6882

Let's just remove it, it doesn't serve a purpose.


oetiker:
yep

PR #792, opened 2017-05-30
Improved buffering

svenpanne:
This is an essential performance patch for rrdcached: Its internal socket buffering scheme had quadratic complexity, basically bringing down rrdcached to its knees for some FETCH commands. We had some use cases where rrdcached shuffled around 130GB(!) internally, although only a few hundred kB were ultimately written out to the socket.

To make the review a bit easier, I made 3 separate commits:

* One purely mechanical change, encapsulating the acces to the write buffer, preparing the way for the next commit.
* The crucial change: Distinguish between size and capacity, and grow the latter geometrically.
* Reap the benefits.

I can squash the changes into a single commit if this is wanted.

### DIFF
--- a/src/rrd_daemon.c
+++ b/src/rrd_daemon.c
@@ -152,6 +152,7 @@ struct listen_socket_s
 
   char *wbuf_data;
   size_t wbuf_size;
+  size_t wbuf_capacity;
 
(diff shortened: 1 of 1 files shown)


oetiker:
thanks

PR #793, opened 2017-05-30
Added SUSPEND/RESUME/SUSPENDALL/RESUMEALL commands.

svenpanne:
Suspending/resuming is essential to make e.g. backups of RRDs while rrdcached is running.


oetiker:
thanks

PR #795, opened 2017-06-11
Unconditionally include <stdint.h>

rubenk:
Code including rrd_client.h should not have to define HAVE_STDINT_H.

I accidentally bumped into this while porting Collectd's build system
from autotools to meson.


rubenk:
Thank you Tobias!

Issue #799, opened 2017-07-01
possible improvements in rrd locking


sourcejedi:
> I guess since writes and reads may not be atomic in any event a
sensible locking for fetch might make sense in any way ...
>
> your patches are welcome ...
>
> cheers
> tobi 

Here is my implementation plan.

I believe it is possible to take a read lock in `rrd_open()`.  Or write lock, depending.  `rrd_lock()` would be deprecated.  For backwards compatibility, the locking could be conditional on a new option to `rrd_open()`.  This would avoid potentially triggering deadlocks in unusual uses of librrd.  At the same time, it would be straightforward to audit a codebase for uses of `rrd_open()` which did not lock.

I've looked at rrdcached, rados, and win32.

I found one issue, with win32.  win32 currently uses non-blocking locks.  The `_locking` function used supports a "blocking" lock where "If the bytes cannot be locked, the program immediately tries again after 1 second. If, after 10 attempts, the bytes cannot be locked, the constant returns an error."  This sounds likely to cause problems.  So I suggest using the `select()` trick for sub-second sleeps in our own loop, e.g. 100 times per second, especially as we already have some winsocket usage.  We can use the same max delay as rados (2 seconds).

This is not an ideal architecture.  On my SheevaPlug (old 1.6ghz ARM), cacti takes ~0.7s to render one graph.  On a page with five graphs, Firefox shows that it fetches the graphs in parallel and they take just under 5 seconds.

So it is probably best for the grapher to opt out of locking, unless someone decides to implement a [seqlock](https://en.wikipedia.org/wiki/Seqlock).  (This might temporarily grab a read lock while checking the mtime, before and after rendering the graph).  Having the grapher opt out will help minimize regressions.

This will still allow `rrdtool dump` to be used for backups without fear of inconsistency.  And implementing this will highlight the locking issue, when developers look at librrd.

Other already-existing issues:

* Readers can block writers indefinitely. E.g. on unix, any user that can read the file can take a standard `fcntl()` advisory read lock.  This would be a Denial of Service on the rrdtool writer.  This is a valid tradeoff, except that it's not documented in RRD.  This should be documented e.g. for applications like smokeping and packages thereof - the current Debian smokeping package leaves these files world-readable. ftp://ftp.sendmail.org/pub/sendmail/LockingAdvisory.txt
[...]

PR #800, opened 2017-07-03
add read locking through rrd_open()

sourcejedi:
> I guess since writes and reads may not be atomic in any event a
sensible locking for fetch might make sense in any way ...
>
> your patches are welcome ...
>
> cheers
> tobi

The implementation is shaped slightly differently to the first plan in #799.  Please disregard the first plan, you can just read the messages for the commits with `RRD_LOCK` in the title.

I have not addressed the 2 already existing issues at the end of the #799 description.

This series is affected by numerous #ifdefs.  I have _not_ tested it beyond `make` on my machine.  If you want to look at the preparatory commits first, that's cool (it should be trivial to rebase though, so I'm presenting this together for initial review).

---

Highlighted excerpt:

So `RRD_LOCK|RRD_READONLY` is intended to take a read lock, however I must
admit to limitations of this new code:
    
* On the Windows _locking() function, read locks are treated the same as
  write locks.
    
  https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/locking
    
* On rados, our write lock expires after 2 seconds.  This write-lock is
  probably useful as a way to detect conflicts.  However I would not want
  to rely on a lock that works like this to guarantee consistent reads for
  a backup.  Fortunately rados implements snapshots, and when we *write* to
  RRDs, we use a single atomic operation.

### DIFF
--- a/src/rrd_open.c
+++ b/src/rrd_open.c
@@ -6,15 +6,21 @@
  * $Id$
  *****************************************************************************/
 
-#include "rrd_tool.h"
-#include "unused.h"
-
 #ifdef WIN32
-#include <stdlib.h>
-#include <fcntl.h>
-#include <sys/stat.h>
+#include <windows.h>
+#if _WIN32_MAXVER >= 0x0602 /* _WIN32_WINNT_WIN8 */
+#include <synchapi.h>
 #endif
 
+#include <stdlib.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <sys/stat.h>
+#include <limits.h>
+#endif /* WIN32 */
+
+#include "rrd_tool.h"
+#include "unused.h"
 
 #ifdef HAVE_BROKEN_MS_ASYNC
 #include <sys/types.h>
@@ -31,8 +37,8 @@
 #define	_LK_UNLCK	0	/* Unlock */
 #define	_LK_LOCK	1	/* Lock */
 #define	_LK_NBLCK	2	/* Non-blocking lock */
-#define	_LK_RLCK	3	/* Lock for read only */
-#define	_LK_NBRLCK	4	/* Non-blocking lock for read only */
+#define	_LK_RLCK	3	/* "Same as _LK_NBLCK" */
+#define	_LK_NBRLCK	4	/* "Same as _LK_LOCK" */
 
 
 #define	LK_UNLCK	_LK_UNLCK
@@ -124,6 +130,9 @@
 #endif
 #endif
 
+static int rrd_rwlock(rrd_file_t *rrd_file, int writelock);
+static int close_and_unlock(int fd);
+
 /* Open a database file, return its header and an open filehandle,
  * positioned to the first cdp in the first rra.
  * In the error path of rrd_open, only rrd_free(&rrd) has to be called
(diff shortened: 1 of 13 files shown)


oetiker:
loking good to me ... and the testsuit seems happy too ! ... thanks!

oetiker:
have you done any performance measurements ?

sourcejedi:
No...

It would be interesting to check

1. the fix for READAHEAD to not use MADV_RANDOM has an effect
2. the fix to make READAHEAD _work_ does not regress from 1)
3. the new madvise() on the RRD header page does not regress IO performance either.

Linux does not implement the drop-behind that I'm worrying about with the RRD header, so my test 3) should not show any improvement.

Regarding lock contention:

I don't have a big setup, I just installed cacti and smokeping on a home server. I'm not sure how to model a real-world heavily-loaded server.  In the first plan I was concerned about holding locks while drawing graphs, but then I learnt the grapher uses rrd_fetch, and so the lock only covers reading the data.

I notice something interesting about rrdcached.  Since readers first request a flush, it seems like readers will always be blocked by writes (which I think is fine), but writes will not tend to be blocked by readers.

There will be some small cpu overhead from additional syscalls (lock, and my extra madvise).

On Windows contended locks will cost extra cpu, from rrdtool waking up 100 times per second to check the lock.  Because I was too lazy to deal with [this](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365203(v=vs.85).aspx) abomination and acquire a Windows build environment to compile-test the result.

sourcejedi:
FIXME: _locking() on Windows says to make sure to unlock before close, because "If a process terminates with a portion of a file locked or closes a file that has outstanding locks, the locks are unlocked by the operating system. However, the time it takes for the operating system to unlock these locks depends upon available system resources. Therefore, it is recommended that your process explicitly unlock all files it has locked when it terminates. If this is not done, access to these files may be denied if the operating system has not yet unlocked them."

oetiker:
@sourcejedi about the fixme  ... are you going to look into this ?

sourcejedi:
Sorry, yes.  I meant I need to fix it, so it shouldn't be merged yet.

sourcejedi:
Do you mind if I rebase this?  The "Fix my Windows locking code" commits being separate would be a pain for anyone who has to bisect this.

oetiker:
sure ... rebase and push --force

oetiker:
I would actually sqash the whole thing ... github lets me do this when merging :)

oetiker:
although at the moment CI is failing so there is another problem with the PR

sourcejedi:
> ERROR: rrdcached@������������������������������������������������������������������������������������������������������正1: Usage: LIST [RECURSIVE] /[<path>]

haha this part isn't even new, but I'm probably going to look at it first :).

src/rrd_open.c
```diff
@@ -599,7 +609,72 @@ int rrd_lock(
...
+
+    return ret;
+}
+#endif
+
+static
+int close_and_unlock(
+    int fd)
+{
+    int ret;
```
sourcejedi: I noticed this ret needs initializing.  Not to spread the blame, but I'm surprised the compiler doesn't warn about it.  It warned me when I had the same problem in rrd_close().

sourcejedi:
> I noticed this ret needs initializing.

Yup, that was the test failure.  And I noticed [yet another oversight](https://github.com/sourcejedi/rrdtool-1.x/commit/2603807b6e03db00156eae7a1a7554d302963801) in my Windows locking.

I've squashed this down, and stripped out the non-interfering fixes for later.

My next task is a PR for the READAHEAD/madvise() fix.  I have a performance test that clearly shows READAHEAD was broken.  However, I need to look into this more.  Because It also showed that my commit which targeted READAHEAD actually regressed a bit - the best performance was caused by a different commit.  The times are slightly non-deterministic, but i can also see an exact difference in the number of "major" page faults.

oetiker:
are you happy with this one too ?

sourcejedi:
Yep, I'd be happy for this to merge now.

oetiker:
thanks!

PR #801, opened 2017-07-05
Fix noise encountered in testing

sourcejedi:
Fix noise encountered in testing #800.
```

OK: recursive list only lists rrd files
ERROR: rrdcached@������������������������������������������������������������������������������������������������������正1: Usage: LIST [RECURSIVE] /[<path>]
OK: escape from cached basedir via symlink denied
```


oetiker:
thanks

PR #802, opened 2017-07-07
Fix madvise (and fadvise)

sourcejedi:
Excerpt:

Passing RRD_READAHEAD to rrd_open() had no effect on current linux.

Test:

```
for i in $(seq 50); do
  dd if=dns.rrd of=x/$i.rrd bs=1M oflag=direct 2>/dev/null
done
/usr/bin/time sh -c 'for i in x/*.rrd; do
  rrdtool dump $i >/dev/null
done'
```

Before:

```
6.52user 0.74system 0:09.46elapsed 76%CPU
(0avgtext+0avgdata 9796maxresident)k
204800inputs+0outputs (25600major+25112minor)pagefaults 0swaps
```

After:

```
5.82user 0.27system 0:06.57elapsed 92%CPU
(0avgtext+0avgdata 9740maxresident)k
204800inputs+0outputs (600major+27020minor)pagefaults 0swaps
```

### DIFF
--- a/src/rrd_open.c
+++ b/src/rrd_open.c
@@ -374,7 +374,11 @@ rrd_file_t *rrd_open(
     }
 #endif
 #if !defined(HAVE_MMAP) && defined(HAVE_POSIX_FADVISE)
-    posix_fadvise(rrd_simple_file->fd, 0, 0, POSIX_FADV_RANDOM);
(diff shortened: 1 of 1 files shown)


oetiker:
wow!

PR #803, opened 2017-07-07
Recently discovered fixes

sourcejedi:
Fixes discovered during PR #800, #801, #802.  (Non-conflicting).

Ideally these would have been included as prep work... but #800 got out of control a bit :).

### DIFF
--- a/src/rrd_cgi.c
+++ b/src/rrd_cgi.c
@@ -754,13 +754,12 @@ static char *includefile(
 
         readfile(filename, &buffer, 0);
         if (rrd_test_error()) {
-            const size_t len = strlen(rrd_get_error()) + DS_NAM_SIZE;
-            char *err = (char *) malloc(len);
-
-            snprintf(err, len, "[ERROR: %s]", rrd_get_error());
+            char err[4096];
+            snprintf(err, sizeof(err), "[ERROR %s]", rrd_get_error());
             rrd_clear_error();
+
             free(buffer);
-            return err;
+            return stralloc(err);
         } else {
             return buffer;
         }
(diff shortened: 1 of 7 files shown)


src/rrd_cgi.c
```diff
@@ -560,13 +559,13 @@ static char *rrdsetenv(
     const char **args)
 {
     if (argc >= 2) {
-        const size_t len = strlen(args[0]) + strlen(args[1]) + 2;
-        char *xyz = (char *) malloc(len);
+        char *xyz = NULL;
 
+        asprintf(&xyz, "%s=%s", args[0], args[1]);
```
oetiker: we have been avoiding asprintf since it is a GNU extension ... 

These functions are GNU extensions, not in C or POSIX.  They are also available under *BSD.  The FreeBSD implementation  sets  strp  to  NULL  on error.

How would folks runing on solaris, aix, hpux handle thins?

sourcejedi: > we have been avoiding asprintf since it is a GNU extension ...

noo, we've been using it and providing a shim, which is the right thing to do.  https://github.com/oetiker/rrdtool-1.x/search?q=asprintf

> The FreeBSD implementation sets strp to NULL on error.

Good for them, but we don't need to rely on that.  FreeBSD still return -1 on failure.

oetiker: yes but that is `rrd_asprintf` so you would have to use that

oetiker: oh and the one in rrd_list.c sliped my notice (new code)

sourcejedi: But we _never_ literally call `rrd_asprintf`, only `asprintf`.  The shim includes some macro magic.  I didn't verify what the magic was doinng, I assumed it did the only sensible thing.

sourcejedi: I see it doesn't... was there a reason this got "mutilated", can we have it back?

614c9e0b8745c57e7d89b93a780e1ed4e128bf92

oetiker: the reason for the existance of the rrd_*print* functions is that they do NOT support locales ...  
normal *print* functions are all locale aware and thus make it hard to produce consistent output.

I did not want to replace the standard *print* functions since we do need locale support in some places ... 

the whole locale thing in this respect is basically about properly outputting numbers ... so here you could easily switch to rrd_asprintf and be done withit.

sourcejedi: noo :(.  I'll leave it to some *other* brave soul to suggest adding a _second_ copy of snprintf.c to the project.

There's some other places that use a stack buffer for formatting error messages, so I copied that instead.

PR #804, opened 2017-07-07
Fix valgrind usage in tests

sourcejedi:
The memory error shown in #801 sneaked past the tests (and hence Travis CI didn't notice it).  Fix the test suite so it actually catches valgrind errors :).

This also caught one memory leak.

### DIFF
--- a/src/rrd_daemon.c
+++ b/src/rrd_daemon.c
@@ -4008,11 +4008,13 @@ static void open_listen_sockets_traditional(void) /* {{{ */
 {
  if (config_listen_address_list_len > 0)
   {
-    for (size_t i = 0; i < config_listen_address_list_len; i++)
+    for (size_t i = 0; i < config_listen_address_list_len; i++) {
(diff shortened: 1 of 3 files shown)


sourcejedi:
Right.  The corrected version shows the grapher is leaky - Travis should fail.

I'm not going to touch the grapher, I think I've finished what I was interested in.

Thanks for the software!  cacti (which uses RRD) is pretty useful to me.

sourcejedi:
In case you ignored Travis due to more recent failures - allow me to te-iterate that this PR breaks CI, as there are unfixed leaks in the grapher.  And maybe more I was glossing over - one leak in xport and one memory error related to list?

I'm not sure you really wanted to merge this PR, unless someone is going to fix or skip the failing tests.

oetiker:
the problem was that valgrind tests started all to fail, so I have to look into this anyway, so I added your patch to at least have them fail sensibly :)

sourcejedi:
Haha :).  Thanks for the update.

Issue #808, opened 2017-07-14
rrdcached pending command fails


stevenh:
The pending command for rrdcached doesn't seem to work, always fails with an error
```sh
pending /test.rrd
-1 No such file or directory
```

The test rrd definitely exists:
```
last /test.rrd
0 1499981700
```

rrd version: 1.6

stevenh:
In addition the docs don't detail anything about what the return format should be which would be nice.

oetiker:
hi stevenh, I just took a look through the code, comparing the two commands (see `rrd_daemon.c` function `handle_request_last` vs `handle_request_pending`) and I don't see how you should be getting such different results except maybe because pending looks only at the cache, and I guess if the file is not in the cache it could respond like this ... make sure to run an update first.

stevenh:
Having looked at the code I believe this is the issue:
https://github.com/oetiker/rrdtool-1.x/blob/master/src/rrd_daemon.c#L1498

oetiker:
### Triage check (master @ 17d1f842)

**Status:** Likely fixed

This was resolved by PR #809 (commit 607cc1b7, "Return success for rrdcached pending on empty"), merged in July 2017. The `handle_request_pending` function in `src/rrd_daemon.c` now returns `RESP_OK` with zero entries when the file is not found in the cache, rather than returning an error. The current code (lines 1560–1585) confirms the fix is in place.

oetiker:
Closing during triage — see the triage analysis above for the specific commit(s) that addressed this. Please reopen with current reproduction steps if you can still reproduce this on rrdtool 1.10.x or current master.

PR #809, opened 2017-07-14
Return success for rrdcached pending on empty

stevenh:
If we have no pending entries for a file return success with no entries instead of returning an error.

### DIFF
--- a/src/rrd_daemon.c
+++ b/src/rrd_daemon.c
@@ -1485,28 +1485,19 @@ static int handle_request_pending(HANDLER_PROTO) /* {{{ */
     return syntax_error(sock,cmd);
 
   file = get_abs_path(pbuffile);
-  if (file == NULL) {
-    rc = send_response(sock, RESP_ERR, "%s\n", rrd_strerror(ENOMEM));
-    goto done;
-  }
+  if (file == NULL)
+    return send_response(sock, RESP_ERR, "%s\n", rrd_strerror(ENOMEM));
 
   pthread_mutex_lock(&cache_lock);
   ci = g_tree_lookup(cache_tree, file);
-  if (ci == NULL)
-  {
-    pthread_mutex_unlock(&cache_lock);
-    rc = send_response(sock, RESP_ERR, "%s\n", rrd_strerror(ENOENT));
-    goto done;
+  if (ci != NULL) {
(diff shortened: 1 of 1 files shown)


stevenh:
This fixes #808

src/rrd_daemon.c
```diff
@@ -1495,7 +1495,7 @@ static int handle_request_pending(HANDLER_PROTO) /* {{{ */
   if (ci == NULL)
   {
     pthread_mutex_unlock(&cache_lock);
-    rc = send_response(sock, RESP_ERR, "%s\n", rrd_strerror(ENOENT));
+    rc = send_response(sock, RESP_OK, "updates pending\n");
     goto done;
   }
```
oetiker: wouldn't it make more sense to just execute the following for loop only IF ci != NULL instead of duplicating the code above ?

oetiker (changes requested):

src/rrd_daemon.c
stevenh: Sure, also cleaned up the rest of the flow as well as the goto isn't needed

src/rrd_daemon.c
```diff
@@ -1485,28 +1485,19 @@ static int handle_request_pending(HANDLER_PROTO) /* {{{ */
...
-  for (size_t i=0; i < ci->values_num; i++)
-    add_response_info(sock, "%s\n", ci->values[i]);
-
-  pthread_mutex_unlock(&cache_lock);
-  rc = send_response(sock, RESP_OK, "updates pending\n");
-done:
   free(file);
-  return rc;
+  pthread_mutex_unlock(&cache_lock);
+  return send_response(sock, RESP_OK, "updates pending\n");
```
oetiker: hmmm still not quite sure if this is correct ... if no pending updates (ci == NULL) are found, then the message 'updates pending' seems odd ...

oetiker (changes requested):

src/rrd_daemon.c
stevenh: The message will actually be:
```
0 updates pending
```
Which IMO makes perfect sense

oetiker: oh my bad ... :) send_response counts what has been added .. all well :) ok

PR #815, opened 2017-08-08
correct the note in the perl binding about the empty filename

flumm:
passing an empty filename to graphv does not work anymore, but the
documentation of the perl-shared binding still said so

correctly note that you have to pass '-'

Signed-off-by: Dominik Csapak <d.csapak@proxmox.com>


oetiker:
correct :) thanks!

PR #820, opened 2017-09-14
Add missing files from win32 to Makefile.am

c72578:
- Missing files were added in the Makefile.am to EXTRA_DIST
- The extra files will be included in the tarball
- See [rrd-users] message:
  https://lists.oetiker.ch/pipermail/rrd-users/2016-August/020443.html


oetiker:
thanks!

PR #821, opened 2017-09-18
Fix error 'optind': undeclared identifier

c72578:
- Fixes compilation error, when building under Windows (e.g. VS2015):
  ..\src\rrd_tune.c(410): error C2065: 'optind': undeclared identifier
- Use options.optind instead of optind
  optind is a member of the struct options


oetiker:
excellent!

PR #822, opened 2017-09-19
Avoid using VLA in optparse.c under Windows (MSC)

c72578:
- Fixes compilation errors, when building under Windows (e.g. VS2015):
  ..\src\optparse.c(195): error C2057: expected constant expression
  ..\src\optparse.c(195): error C2466: cannot allocate an array of
  constant size 0
  ..\src\optparse.c(195): error C2133: 'optstring': unknown size
- Variable length arrays are not currently supported in Visual Studio
- Use malloc() instead of variable length array in case of MSC


oetiker:
thanks

PR #823, opened 2017-09-20
Include unistd.h only if HAVE_UNISTD_H is defined

c72578:
- Fixes compilation error, when building under Windows (e.g. VS2015):
  ..\src\rrd_create.c(18): fatal error C1083: Cannot open include file:
  'unistd.h': No such file or directory


oetiker:
Hi Wolfgang, once you are through, will you publish new windows rrd binaries ?

c72578:
Getting closer :smile:

PR #824, opened 2017-09-23
Add detailed PDP calculation logics explanation

diphuaji:
Detailed PDP calculation logics explanation with an example. 
It provides an easy-to-understand explanation of how PDPs are produced, so new users can easily justify calculation results themselves.


diphuaji:
What's wrong with the travis...
I only edited a pod file.

oetiker:
thanks!

PR #832, opened 2017-10-09
Fix formal parameter different from declaration

c72578:
- Fixes compiler warnings C4028 (MSVC):
  formal parameter 2 different from declaration src\rrd_first.c 87
  formal parameter 3 different from declaration src\rrd_graph_helper.c 1500
  formal parameter 5 different from declaration src\rrd_modify.c 90


oetiker:
thanks

PR #839, opened 2017-11-17
Define HAVE_G_REGEX_NEW, remove PCRE deps (MSVC)

c72578:
- Define HAVE_G_REGEX_NEW 1 in rrd_config.h
  glib has g_regex_new since 2.14
- Remove dependency on PCRE, which is not required any more


oetiker:
thanks

PR #845, opened 2017-11-28
Remove fdatasync, fsync from rrd_config_bottom.h

c72578:
- fdatasync() has been removed since RRDtool v1.4.0.
  See commit: fc968dd
- fixes "Can't compile with without fsync and fdatasync" error
  on systems, where HAVE_FDATASYNC or HAVE_FSYNC are not defined


oetiker:
thanks

Issue #865, opened 2018-01-26
threading issues in python bindings


matwey:
Hello,

I use python bindings and rrdtool 1.7.0 and I see the following when using multiple threads:

```python
import rrdtool
import shlex
from threading import Thread

def render(command, format):
        format = format.upper()
        args = ['--imgformat',format] + shlex.split(command)
        graph = rrdtool.graphv('-', *args)
        return graph['image']

def run():
        command = "-b 1024 -v Bytes -l 0 DEF:free_avg=db.rrd:value:AVERAGE DEF:used_avg=db.rrd:value:AVERAGE AREA:free_avg#ccffcc:Free:STACK GPRINT:free_avg:AVERAGE:\"%5.1lf%sB\\l\" AREA:used_avg#ffddcc:Used:STACK GPRINT:used_avg:AVERAGE:\"%5.1lf%sB\\l\""

        for x in range(0, 1000):
                render(command, 'SVG')

threads = 4

for x in range(0, threads):
        t = Thread(target = run)
        t.start()
```

```
Exception in thread Thread-4:
Traceback (most recent call last):
  File "/usr/lib64/python3.4/threading.py", line 911, in _bootstrap_inner
    self.run()
  File "/usr/lib64/python3.4/threading.py", line 859, in run
    self._target(*self._args, **self._kwargs)
  File "./test.py", line 17, in run
    render(command, 'SVG')
  File "./test.py", line 10, in render
    graph = rrdtool.graphv('-', *args)
rrdtool.OperationalError: '1' is not a valid function name in 1

*** Error in `python': malloc(): smallbin double linked list corrupted: 0x00007efc74048b30 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x721af)[0x7efc83a8a1af]
/lib64/libc.so.6(+0x77706)[0x7efc83a8f706]
/lib64/libc.so.6(+0x79bd5)[0x7efc83a91bd5]
/lib64/libc.so.6(+0x7a5ac)[0x7efc83a925ac]
/lib64/libc.so.6(realloc+0x12f)[0x7efc83a93b5f]
/usr/lib64/librrd.so.8(+0x21f2d)[0x7efc82ed4f2d]
/usr/lib64/librrd.so.8(+0x220e4)[0x7efc82ed50e4]
/usr/lib64/librrd.so.8(+0x22354)[0x7efc82ed5354]
[...]

PR #866, opened 2018-01-27
Fix/python race

matwey:
Hello,

This is partial fix for #865 

Currently, I don't see python crashing anymore. But now I see warnings from GLibc, which I believe produced by librrd itself, not bindings. 

```
(process:4793): Pango-WARNING **: failed to create cairo scaled font, expect ugly output. the offending font is 'DejaVu Sans Mono 8'

(process:4793): Pango-WARNING **: font_face status is: no error has occurred

(process:4793): Pango-WARNING **: scaled_font status is: invalid matrix (not invertible)

(process:4793): Pango-WARNING **: shaping failure, expect ugly output. shape-engine='PangoFcShapeEngine', font='DejaVu Sans Mono 8', text='Bytes'

(process:4793): GLib-WARNING **: corrupted double-linked list detected
```

### DIFF
--- a/bindings/python/rrdtoolmodule.c
+++ b/bindings/python/rrdtoolmodule.c
@@ -64,9 +64,6 @@ static const char *_version = "0.1.10";
 static PyObject *rrdtool_OperationalError;
 static PyObject *rrdtool_ProgrammingError;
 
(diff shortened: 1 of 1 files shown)


oetiker:
thanks!

PR #875, opened 2018-02-13
Do not use static variable in gfx_prep_text()

matwey:
Move last_tabwidth to image_desc_t

Signed-off-by: Matwey V. Kornilov <matwey.kornilov@gmail.com>


oetiker:
thanks!

PR #879, opened 2018-03-06
Fix rrdtool.graph python binding

Tomo59:
It was failing with 'munmap_chunk(): invalid pointer' error in python 3.6.3


oetiker:
thanks

PR #881, opened 2018-03-25
Update MinGW-w64 build instructions

c72578:
- Info added, where to find compiled binaries:
  src/.libs/
- Added perl-Pod-Html to Fedora dependencies
- Added base-devel to MSYS2 build instructions


oetiker:
thanks!

PR #887, opened 2018-05-25
Added CSV format to rrdtool dump

fbsanchez:
Hello!

I just added CSV format output to rrdtool dump utility, it also is able to select the desired CF to be exported (max,avg,min,last)

Hope you find it useful

Thank you!

### DIFF
--- a/src/rrd.h
+++ b/src/rrd.h
@@ -297,7 +297,8 @@ struct rrd_t;
     int rrd_dump_opt_r(const char *filename,
                        char *outname,
                        int opt_noheader,
-                       int opt_mode);
+                       int opt_mode,
+                       char *out_cf);
     int       rrd_dump_r(
     const char *filename,
     char *outname);
@@ -316,7 +317,8 @@ struct rrd_t;
     int opt_header,
     rrd_output_callback_t cb,
     void *user,
-    int opt_mode);
+    int opt_mode,
+    char *opt_cf);
 
 /* Transplanted from rrd_parsetime.h */
     typedef enum {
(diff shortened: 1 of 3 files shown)


oetiker:
hmmm adding such a function to dump seems a bit odd ... I would have expected it to show up in xport for example ... what is your thinking on this ?

fbsanchez:
Hello oetiker, maybe the xport is the right place, I use dump because the default XML format is not useful for my purposes :) I've no time right now to change the placement, is this placement a handicap?

oetiker:
yes,because except in caes where you want to dump the entire rrd file (to xml) you would use either rrdtool fetch or rrdtool xport to extract data

fbsanchez:
Hello! I'm gonna move the changes to the export functionality, I'll use this PR :) thank you

fbsanchez:
Uhm, while checking the xport feature I'm not fully sure if it covers all my needs, as you commented: 

> except in case where you want to dump the entire rrd file (to xml)

is just that case, I want to retrieve entire content of the RRD file (only data) CSV sepparated, utimestamp + CDs

I think our desired functionality could be useful in dump, because we need no time range to export, since we need entire RRD content (filtered by aggregation origin - AVERAGE/MIN/MAX...)

Do you think this behaviour is useful for your SW?

oetiker:
the problem is that dump returns structured data ... suitable for restore ... you might be better served with 'rrdtool fetch' if you desire 'original' data

netniV:
Personally, I think it would be advisable to have it in both places.  Obviously, CSV formatted data does not need to be used for restore, but could be used with other systems for importing (such as mysql, excel, etc).  Additionally, if JSON is handled by both dump/xport, I think having CSV in both places also makes sense.

PR #893, opened 2018-07-13
Define HAVE_RRD_RESTORE (MSVC)

c72578:
- Enables restore functionality on Windows MSVC builds
- Fixes ERROR: the instance of rrdtool has been compiled without XML
  import functions


oetiker:
thanks!

PR #897, opened 2018-07-27
Allow multiple lines in graph titles

netniV:
This PR contains work to allow multiple lines to be used within the graph titles.  It current assumes a question mark as the character to split by.  Further work would offer a parameter to customise the split character or even recognise <br> and \n.


oetiker:
Hi @netniV, thanks for your PR ... yes please make the new line character to be "\n" (literally backslash n). The question mark already has a well defined job (as a question mark)

oetiker:
also please to not commit changes to Makefile.in and configure

netniV:
Yeah someone else pointed out to me that there were makefile changes.  I think I made a mistake with the commit and added the "-a" as I do with other projects without realising.  Would it be worth including the <br> since it is a field that supposedly takes tags?

netniV:
I'm not sure how a program will pass a \n in the title but made the change you've requested and you should now see a cleaner difference.

src/rrd_graph.c
```diff
@@ -6147,3 +6155,27 @@ void time_clean(
...
     result[jj] = '\0'; /* We must force the end of the string */
 }
+
+image_title_t graph_title_split(
+    const char *title)
+{
+    image_title_t retval;
+    char *str;
+    int count = 0;
+    const char delim[2] = "?";
```
oetiker: the delimiter can not be a `?`  make it a litteral `\n`

src/rrd_graph.c
```diff
@@ -6162,7 +6162,7 @@ image_title_t graph_title_split(
     image_title_t retval;
     char *str;
     int count = 0;
-    const char delim[2] = "?";
+    const char delim[2] = "\n";
```
oetiker: you could make your code handle both ... a chr(10) as well as a literal '\n' in that way it would work regardless if there is an interpreter in the path that already evaluates \n or not.

netniV: Oh I see what you mean.  In the same way you'd format a printf string.

netniV:
OK, hopefully this time I got it right and made you a happy bunny ;-)

src/rrd_graph.c
```diff
@@ -6147,3 +6155,75 @@ void time_clean(
...
+    int start = 0;
+    int length = 0;
+    int pos = 0;
+    int count = 0;
+    int delim = 1;
+
+    retval.lines = malloc((MAX_IMAGE_TITLE_LINES + 1 ) * sizeof(char *));
+    length = strlen(title);
+
+    char *delims[6] = { "\n", "\r", "\\n", "\\r", "<br>", "<br/>" };
```
oetiker: there is already the option to use simple markup in rrdtool strings ... so in that way <br> is a good idea ... bringing \r into the mix opens a can of worms I would rather not want to touch ... please stick to \n since this is what most languages will use for breaking a line ...

netniV: OK, I'll remove \r and try to find this rrdgraph.pod (whatever/wherever that may be 👍 )

netniV: Got it, in /docs

netniV: OK this has now been done.

oetiker:
great! thanks!

netniV:
Quick question, what is the normal release cycle of changes like these?  Just wondering how long it would take to appear in packages if people ask me 👍

oetiker:
well I will be doing 1.8 release pretty soon, but to see this in packages you will ususally have to wait a few months/years ... many people still run 1.4

netniV:
I think when it comes to running with Cacti, we've seen many using 1.3 purely because their running Cacti 0.8.8 and their systems are so old, no new packages are being made.  Those running 1.x seem to be 1.5 or higher but not many are running 1.7 yet.

PR #901, opened 2018-08-16
lua: don't include rrd_restore if it wasn't included

karlp:
The lua module compiles ok, but fails to load.

> rrd=require("rrd")
error loading module 'rrd' from file '/usr/lib/lua/rrd.so':
	Error relocating /usr/lib/lua/rrd.so: rrd_restore: symbol not found
stack traceback:
	[C]: ?
	[C]: in function 'require'

Signed-off-by: Karl Palsson <karlp@etactica.com>


oetiker:
thanks!

PR #907, opened 2018-09-26
Update perl-shared README for MSVC and ActivePerl

c72578:
- Using MSVC, the last supported ActivePerl version is 5.16,
  e.g. ActivePerl-5.16.3.1604-MSWin32-x86-298023.msi
- Since 5.18, ActivePerl is compiled with GCC compilers, and Microsoft
  toolchains are *not* compatible
- See this FAQ for further details:
  https://community.activestate.com/faq/windows-compilers-perl-modules


oetiker:
thanks

Issue #914, opened 2018-10-30
rrdtool graph segfaults when --title is used


az143:
the current code from git (head at ecf80e6729)
segfaults when trying to graph anything with a delimiter-free --title, e.g. --title "hello".

to reproduce: take any plain rrd file (example here test.rrd),
then try rrdtool graph /tmp/testme.png --title "hello" "DEF:health=test.rrd:health:AVERAGE" AREA:health#CD5C5C:Health

you'll get spurious test output  on stdout
title: hello
followed by a segmentation fault.

gdb tells me that strlen() inside of graph_title_split() is what fails. it looks as if
the code in question (from issue #897, commit cd4a72856) allocates an array of six possible delimiters, but populates only four and then tries to walk all six.

changing the delimiter walking loop from 6 to just the 4 present ones
            for(int i=0; i<4;i++) {
gets rid of the segfault. 

(the output is semi-broken in terms of title lines: i get multiple lines of gunk despite requesting just
one.)

PR #915, opened 2018-10-30
fix for #914, rrdtool graph with --title segfaults

az143:
i've updated the graph_title_split() function a bit to not segfault and to handle various corner cases better (e.g. single title w/o delimiter didn't work before, back-to-back delimiters or leading or trailing ones).

### DIFF
--- a/src/rrd_graph.c
+++ b/src/rrd_graph.c
@@ -6195,7 +6195,7 @@ image_title_t graph_title_split(
         delim = 0;
 
         while (delim == 0 && start + pos < length) {
-            for(int i=0; i<6;i++) {
+            for(int i=0; i<4;i++) {
                 int delim_size = strlen(delims[i]);
                 int delim_match = strncasecmp(title+start+pos, delims[i], delim_size);
 
--- a/src/rrd_graph.c
+++ b/src/rrd_graph.c
@@ -6178,50 +6178,61 @@ image_title_t graph_title_split(
     const char *title)
 {
     image_title_t retval;
-    int start = 0;
-    int length = 0;
-    int pos = 0;
-    int count = 0;
-    int delim = 1;
+    int count = 0;              /* line count */
+    int found_delim;
 
     retval.lines = malloc((MAX_IMAGE_TITLE_LINES + 1 ) * sizeof(char *));
-    length = strlen(title);
 
-    char *delims[6] = { "\n", "\\n", "<br>", "<br/>" };
-    printf("title: %s\n", title);
+    char *delims[4] = { "\n", "\\n", "<br>", "<br/>" };
+    // printf("unsplit title: %s\n", title);
+
+    char *consumed = strdup(title); /* allocates copy */
     do
     {
-        pos = 0;
-        delim = 0;
+       /*
+         search for next delimiter occurrence in the title,
+         if found, save the string before the delimiter for the next line & search the remainder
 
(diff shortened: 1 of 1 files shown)


src/rrd_graph.c
```diff
@@ -6178,50 +6178,61 @@ image_title_t graph_title_split(
...
-    int delim = 1;
+    int count = 0;              /* line count */
+    int found_delim;
 
     retval.lines = malloc((MAX_IMAGE_TITLE_LINES + 1 ) * sizeof(char *));
-    length = strlen(title);
 
-    char *delims[6] = { "\n", "\\n", "<br>", "<br/>" };
-    printf("title: %s\n", title);
+    char *delims[4] = { "\n", "\\n", "<br>", "<br/>" };
```
oetiker:
```C
char *delims[] = { "\n", "\\n", "<br>", "<br/>" };
```

would be more robust

src/rrd_graph.c
```diff
@@ -6178,50 +6178,61 @@ image_title_t graph_title_split(
...
-        }
+       /*
+         search for next delimiter occurrence in the title,
+         if found, save the string before the delimiter for the next line & search the remainder
+
+         note: this logic cannot work if the title contains different delimiters,
+         as there's no guarantee that we'll find the earliest delimiter first!
+       */
+       found_delim = 0;
+       for(unsigned int i=0; i < 4; i++)
```
oetiker:
```C
for(unsigned int i=0; i < sizeof(delims); i++)
```

az143:
indeed, that's better - i had just opted for the minimum amount of changes that resolve the underlying issue.
anyway, both of your requested changes have been made.

oetiker:
thanks!

netniV:
As noted, this patch doesn't handle multiple delimiters and I verified that looking at the code.  

Whilst we would hope that a calling program will always use a single type, there are situations such as with Cacti, where multiple may be provided.  This can happen because the admin may entire a title style, but a device may also return a delimited in a value that gets used in a dynamic title.  

If the style picked by the device uses a delimiter earlier in the list, it would get broken up different and the string may include the operators intended break a line leading to unexpected output.  

Handling this corner case can be done with only a small improvement on what has been done here by I think:

1. Search for the position of each delimiter
2. Find the lowest index and substring from the current position to that location
3. Search again but just for the delimiter we have just used since we already have the lowest of the rest
4. Repeat part 2 until none found or the maximum lines is reached

If you feel it's worth it, Either I or @az143 (if they feel up to it) could apply that change.

oetiker:
improvements are always welcome :) since we are supporting multiple line break chars, we might as well allow a mix ...

netniV:
I'll take a look at later this week then unless @az143 feels he would rather do it.

az143:
On Wed, 07 Nov 2018 15:33:41 -0800, Mark Brugnoli-Vinten writes:
>I'll take a look at later this week then unless @az143 feels he would
>rather do it.

please go for it; i'm a bit busy at the moment.

regards
az


-- 
Alexander Zangerl + GPG Key 2FCCF66BB963BD5F + http://snafu.priv.at/
"You are SUCH a useless use of a constant." -- flannelcat

netniV:
@az143 and @oetiker, I have created a PR with what I believe should be working.  Feel free to review it and let me know.

PR #917, opened 2018-11-15
Update to PR #915

netniV:
I believe this new code should handle the multiple delimiter split better as discussed in #914

### DIFF
--- a/src/rrd_graph.c
+++ b/src/rrd_graph.c
@@ -6179,60 +6179,64 @@ image_title_t graph_title_split(
 {
     image_title_t retval;
     int count = 0;              /* line count */
-    int found_delim;
+    char *found_pos;
+    int found_size;
 
     retval.lines = malloc((MAX_IMAGE_TITLE_LINES + 1 ) * sizeof(char *));
 
     char *delims[] = { "\n", "\\n", "<br>", "<br/>" };
+
     // printf("unsplit title: %s\n", title);
 
     char *consumed = strdup(title); /* allocates copy */
     do
     {
-       /*
-         search for next delimiter occurrence in the title,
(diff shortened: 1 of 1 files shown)


oetiker:
thanks!

c72578:
gcc shows the following warning:
<pre>
rrd_graph.c: In function ‘graph_title_split’:
rrd_graph.c:6203:43: warning: iteration 4 invokes undefined behavior [-Waggressive-loop-optimizations]
             int delim_size = strlen(delims[i]);
                                     ~~~~~~^~~
rrd_graph.c:6200:9: note: within this loop
         for(unsigned int i=0; i < sizeof(delims); i++)
         ^~~
</pre>

c72578:
`sizeof(delims)` determines the size of delims in bytes, which can be n=16 (4x4) on 32-bit and n=32 (4x8) on 64-bit systems. So we need:
`for(unsigned int i=0; i < sizeof(delims) / sizeof(delims[0]); i++)`
instead of
`for(unsigned int i=0; i < sizeof(delims); i++)`

I will submit a PR.

netniV:
Ah, yes.  Please do.

c72578:
PR created: https://github.com/oetiker/rrdtool-1.x/pull/919

oetiker:
would that fail it the delim strings were of different size ?

c72578:
@netniV Have you got a minimal rrdtool example by hand for testing the multiple delimiter split? I have just been testing, but did not get the expected results yet.

netniV:
@c72578 Thanks! I was going to post that I use Cacti to generate my tests but I could have got the command lines.  Unfortunately, day job got in the way and you'd resolved it before I got back to you 👍

PR #919, opened 2018-11-23
Fix loop, sizeof(delims)

c72578:
- Update to https://github.com/oetiker/rrdtool-1.x/commit/2154d02

### DIFF
--- a/src/rrd_graph.c
+++ b/src/rrd_graph.c
@@ -6197,7 +6197,7 @@ image_title_t graph_title_split(
         */
         found_pos = 0;
         found_size  = 0;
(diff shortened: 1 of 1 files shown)


src/rrd_graph.c
```diff
@@ -6197,7 +6197,7 @@ image_title_t graph_title_split(
         */
         found_pos = 0;
         found_size  = 0;
-        for(unsigned int i=0; i < sizeof(delims); i++)
+        for(unsigned int i=0; i < sizeof(delims) / sizeof(delims[0]); i++)
```
oetiker: so sizeof(delims) does not return the length of the array ? what does it return ?

oetiker:
thanks

PR #921, opened 2018-11-29
Fix tests to support Windows time zone name

c72578:
- On Windows, names of times zones are different.
  Substitute 'W. Europe Standard Time' with 'CET' by function seddif
- Removes unnecessary diff output and fixes therefore failing tests
  under Windows


oetiker:
thanks!

PR #923, opened 2018-12-14
Sync librrd-4.def with librrd.sym

c72578:
- Export the same functions in the dll from MSVC builds under Windows
  as listed in librrd.sym


oetiker:
thanks

PR #924, opened 2018-12-18
Check for gobject-2.0 in configure

c72578:
- Since pango version 1.43.0, -lgobject-2.0 is not added automatically
  any more during configure. Remark: pango.pc uses Requires.private
  now for gobject-2.0 instead of formerly Requires: gobject-2.0
- Check for gobject-2.0 in configure to make sure, that -lgobject-2.0
  is added to ALL_LIBS
- Fixes the following linker errors:
<pre>.libs/librrd_la-rrd_graph.o: In function `im_free':
src/rrd_graph.c:465: undefined reference to `g_object_unref'
.libs/librrd_la-rrd_graph.o: In function `rrd_graph_init':
src/rrd_graph.c:4840: undefined reference to `g_object_unref'
</pre>


oetiker:
thanks

PR #925, opened 2018-12-18
Add -lws2_32 for MinGW builds by configure

c72578:
- Add -lws2_32 to LIBS in case of MinGW or MinGW-w64 builds.
  This makes building Windows executables more straightforward using
  configure.
- Update README-MinGW-w64, remove LIBS='-lws2_32' from configure
  instructions.


oetiker:
cool ... this is all about cross compilation, right ? I guess for the next release I should build windows binaries. did you get the perl module to work too like that ?

Issue #913, opened 2018-10-23
Inconsistency in time for values


iuyiuy:
There seems to be an inconsistency with which value is displayed at what time between the xport and dump commands. The times and values match in 1.4.8, but not in 1.5.0, 1.6.0 or the newest 1.7.0 from the repository.

Here are some examples using the same rrd file for 3 of the versions. Note: Newer versions need the -t/--showtime argument to show the time. And i've censored the e-mail, for the sake of online crawlers for spam.

1.4.8:
```
kbj@kbj-dev ~/git $ rrdtool | head -n1
RRDtool 1.4.8  Copyright 1997-2013 by Tobias Oetiker <***@***>
kbj@kbj-dev ~/git $ rrdtool xport --start 1540207800 --end 1540208700 DEF:ns10=/tmp/ns10.rrd:ns10:AVERAGE XPORT:ns10 | grep '<row>'
    <row><t>1540208100</t><v>9.3745082714e-01</v></row>
    <row><t>1540208400</t><v>1.8551453360e+02</v></row>
    <row><t>1540208700</t><v>NaN</v></row>
    <row><t>1540209000</t><v>NaN</v></row>
kbj@kbj-dev ~/git $ rrdtool dump /tmp/ns10.rrd | grep '2018-10-22' | grep '<row>' | sort -n | uniq | tail -n3
			<!-- 2018-10-22 13:30:00 CEST / 1540207800 --> <row><v>3.5072563673e+01</v></row>
			<!-- 2018-10-22 13:35:00 CEST / 1540208100 --> <row><v>9.3745082714e-01</v></row>
			<!-- 2018-10-22 13:40:00 CEST / 1540208400 --> <row><v>1.8551453360e+02</v></row>
```

1.6.0:
```
kbj@kbj-dev2:~/git$ rrdtool | head -n1
RRDtool 1.6.0  Copyright by Tobias Oetiker <***@***>
kbj@kbj-dev2:~/git$ rrdtool xport -t --start 1540207800 --end 1540208700 DEF:ns10=/tmp/ns10.rrd:ns10:AVERAGE XPORT:ns10 | grep '<row>'
    <row><t>1540207800</t><v>9.3745082714e-01</v></row>
    <row><t>1540208100</t><v>1.8551453360e+02</v></row>
    <row><t>1540208400</t><v>NaN</v></row>
[...]

PR #928, opened 2019-01-03
Fix off by one step error reported in  issue #913

oetiker:


oetiker:
resolves #913

pmorch:
Thanks Tobi, this fixes the issue we were having.

But this only fixes the `rrdtool xport` part of #913, right? Now `rrdtool xport` and `rrdtool graph` disagree. The `rrdtool graph` part of the fix still needs fixing, right?

Or should a new issue be created for the similar `rrdtool graph` issue? (Again, we don't actually need it)

oetiker:
what is the problem with graph ?

PR #931, opened 2019-01-03
Added some documentation for the rrdxport output format - see #913

pmorch:
More or less what we discussed in #913

**About the link to PDP_calculation_explanation.html**

You mentioned https://github.com/oetiker/rrdtool-1.x/blob/master doc/PDP_calculation_explanation.pod in #913 . While writing the documentation I decided that adding a link to that document from rrdxport.pod was the way to go.

However that is a long link. While generating rrdxport.txt from rrdxport.1, I got this output:

    » make
      GEN      rrdxport.1
      GEN      rrdxport.txt
    rrdxport.1:290: warning [p 3, 2.0i]: can't break line
      GEN      rrdxport.html

And I really dislike warnings. The problem is that line 290 is that long URL. So I had these alternatives:

1. Add PDP_calculation_explanation.pod to the POD definition in `Makefile.in` so  it is also generated, and then include a relative link to  ./PDP_calculation_explanation instead of the full absolute link to  github.com.
1. Fix the warning by adding `-Wbreak` to the `@NROFF@` command in Makefile.in
1. Omit the link entirely
1. Live with the warning

I went with the first option, also because having links to a hardcoded domain like github.com is asking for future bitrot, and there aren't any other links to github.com (yet). You may decide otherwise - I just wanted to be explicit about my choice here.

I then discovered that PDP_calculation_explanation.pod itself had perlpod errors that I fixed and that it also generated the "can't break line" warning (and so did many other .pod files). So I added `-Wbreak` to the `@NROFF@` command in Makefile.in to disable all these warnings.


doc/rrdxport.pod
```diff
@@ -141,6 +145,33 @@ The resulting data section is:
...
+be the first sample starting at or immediately after B<--start>. The last
+sample will be the one ending at or immediately after B<--end>.
+
+Each sample has a timestamp and one or more values. The timestamps associated
+with a value in rrdtool ALWAYS represent the time the sample was taken. Since
+any value you sample will represent some sort of past state your sampling
+apparatus has gathered, the timestamp will always be at the end of the sampling
+period.
+
+Rrdtool does not store the actual samples, but does internal resampling of the
```
c72578: Rrdtool -> RRDtool

doc/rrdxport.pod
```diff
@@ -141,6 +145,33 @@ The resulting data section is:
...
+values presented to it. Nevertheless when a data value is presented with a
+single timestamp the timestamp is at the end of the period the value
+represents. Note that the timestamp itself is outside the period the sample is
+valid for. For more details about this, see L<PDP calculation
+explanation|PDP_calculation_explanation>.
+
+So the time range for a sample with a timestamp is actually
+C<< [ timestamp - step ; timestamp [ >>.
```
c72578: timestamp [
->
timestamp ]

doc/rrdxport.pod
pmorch: This was cut'n'pasted verbatim from [Tobi's comment](https://github.com/oetiker/rrdtool-1.x/issues/913#issuecomment-450800212), but I can see RRDtool is what is used also on the [home page](https://oss.oetiker.ch/rrdtool/). I'll fix that where you mention it and also in the paragraph above it.

oetiker:
thanks!

c72578:
The changes concerning rrd_pdpcalc.pod should be made in Makefile.am instead (in addition) to Makefile.in
Otherwise bootstrap or autoreconf will overwrite the changes.

oetiker:
oops indeed did not spot that!

pmorch:
I've just pushed that change too to my branch.

oetiker:
your change should have gone into `doc/Makefile.am` and not `doc/Makefile.in`

c72578:
@pmorch: I have added your commit https://github.com/pmorch/rrdtool-1.x/commit/9c6a465 to a new PR https://github.com/oetiker/rrdtool-1.x/pull/936 and also added -Wbreak to doc/Makefile.am

PR #932, opened 2019-01-03
Fixes for .pod files

pmorch:
While creating #931 I noticed that `podchecker *.pod`  returned a number of errors and warnings. I also noticed that the links to AT-STYLE TIME SPECIFICATION could include the `#AT-STYLE-TIME-SPECIFICATION` in the URL, so I fixed these things.


doc/rrdinfo.pod
```diff
@@ -5,7 +5,8 @@ rrdinfo - extract header information from an RRD
 =head1 SYNOPSIS
 
 B<rrdtool> B<info> I<filename>
-S<[B<--daemon>|B<-d> I<address> S<[B<--noflush>|B<-F>]>]>
+
```
c72578: This additional line would lead to:
<pre>
rrdtool info filename

[--daemon|-d address [--noflush|-F]]
<pre>

doc/rrdinfo.pod
```diff
@@ -5,7 +5,8 @@ rrdinfo - extract header information from an RRD
...
+S<[B<--daemon>|B<-d> I<address> [B<--noflush>|B<-F>]>]>
```
c72578: There is one '>' too much.
See: https://github.com/oetiker/rrdtool-1.x/blob/a9731c5243d73124711b8d43556a9ca0d6a6245d/doc/rrdinfo.pod
Should be:
`S<[B<--daemon>|B<-d> I<address> [B<--noflush>|B<-F>]]>`

pmorch: Yes, you're right. I thought I counted them. Strange podchecker didn't catch that. I can see how it got rendered in HTML that an extra `>` was present in the output. Thanks.

oetiker:
thank you very much!

PR #936, opened 2019-01-09
Update doc/Makefile.am

c72578:
- Add commit from pmorch: Include rrd_pdpcalc.pod in Makefile.am
- Add -Wbreak, as in:
  https://github.com/oetiker/rrdtool-1.x/commit/564aed3


oetiker:
thanks

PR #937, opened 2019-01-10
Remove function substring from rrd_graph.c

c72578:
- The function substring is not used any more. This was a leftover
  from https://github.com/oetiker/rrdtool-1.x/commit/cd4a728
- Fixes rrd_graph.c:6159:7: warning:
  no previous prototype for 'substring' [-Wmissing-prototypes]


oetiker:
thanks!

PR #938, opened 2019-01-10
Use g_list_free_full instead of g_list_foreach

c72578:
- Fixes the following compiler warnings in rrd_create.c:
<pre>
  236:34: warning: cast between incompatible function types from
  'void (*)(void *)' to 'void (*)(void *, void *)'
  [-Wcast-function-type] g_list_foreach( sources, (GFunc)free, NULL );

  1022:44: warning: cast between incompatible function types from
  'void (*)(rrd_file_t *)' {aka 'void (*)(struct rrd_file_t *)'} to
  'void (*)(void *, void *)' [-Wcast-function-type]
  g_list_foreach(sources_rrd_files, (GFunc) cleanup_source_file, NULL);
</pre>

### DIFF
--- a/src/rrd_create.c
+++ b/src/rrd_create.c
@@ -233,8 +233,7 @@ done:
     }
     if (sources != NULL) {
         // this will free the list elements as well
-        g_list_foreach( sources, (GFunc)free, NULL );
-        g_list_free( sources );
+        g_list_free_full ( sources, (GDestroyNotify) free );
         sources = NULL;
     }
     if (template != NULL) {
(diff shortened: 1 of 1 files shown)


oetiker:
thanks

c72578:
Remark: `g_list_free_full` requires glib 2.28 or newer, see:
https://developer.gnome.org/glib/stable/glib-Doubly-Linked-Lists.html#g-list-free-full

oetiker:
I think requiring a 7 year old library is ok :)

PR #940, opened 2019-01-13
Re-enable 0-width lines

### LINKED ISSUE #930: "Line0" is not processed any more
Hi

The original bug report is at https://bugs.debian.org/891491

Commit 1ace836524e58f4db3c30e7dea0e11207ac451f5 changed the LINE width sanity test:
```
                if (gdp->linewidth < 0) {
                    rrd_set_error
                        ("LINE width '%s' is less than 0 in line '%s'\n",
                         &line[*eaten], line);
                    return 1;
                }
```
into
```
      if ((getDouble(t,&linewidth,&x))||(linewidth<=0)) {
	rrd_set_error("Bad line width: %s",t); return NULL;
      }
```
so that LINE0 doesn't work anymore in 1.7.

I'm not quite sure how to handle that bug report. I suppose I should answer that LINE0 is no longer supported, and that's it?  Or would you consider re-adding 0-width support for compatibility?
In the former case, adding a note in the CHANGES would be nice.
In the later case, changing `linewidth<=0` into `linewidth<0` would probably do the trick.

nirgal:
Some people were using these invisible lines as a base for stacking other stuff.

This fixes #930

### DIFF
--- a/src/rrd_graph_helper.c
+++ b/src/rrd_graph_helper.c
@@ -545,7 +545,7 @@ static graph_desc_t* newGraphDescription(image_desc_t *const im,enum gf_en gf,pa
     double linewidth = 1;
     char *t,*x;
     if ((t=getKeyValueArgument("linewidth",1,pa))&&(*t!=0)) {
-      if ((getDouble(t,&linewidth,&x))||(linewidth<=0)) {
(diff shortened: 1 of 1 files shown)


oetiker:
thanks

PR #941, opened 2019-01-15
Define realpath for MinGW builds

c72578:
- Define realpath the same way as in win32/rrd_config.h
- Fixes the following compiler warnings in rrd_client.c:
<pre>
  115:11: warning: implicit declaration of function 'realpath';
  [-Wimplicit-function-declaration] ret = realpath(path, NULL);
  115:11: warning: nested extern declaration of 'realpath'
  [-Wnested-externs]
  115:9: warning: assignment to 'char *' from 'int' makes pointer from
  integer without a cast [-Wint-conversion] ret = realpath(path, NULL);
  132:22: warning: assignment to 'char *' from 'int' makes pointer from
  integer without a cast [-Wint-conversion]
  dir_path = realpath(dir, NULL);
</pre>

### DIFF
--- a/src/rrd_client.c
+++ b/src/rrd_client.c
@@ -38,6 +38,10 @@
 
 #endif
 
+#ifdef __MINGW32__
+#define realpath(N,R) _fullpath((R),(N),_MAX_PATH)
+#endif
+
 #include "rrd_strtod.h"
 #include "rrd.h"
 #include "rrd_tool.h"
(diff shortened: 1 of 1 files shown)


c72578:
Further information concerning realpath under MinGW:
https://sourceforge.net/p/mingw/bugs/2207/
https://sourceforge.net/p/mingw/patches/256/#ef43

Used the suggested simple solution here:
`#define realpath(N,R) _fullpath((R),(N),_MAX_PATH)`

PR #942, opened 2019-01-16
Remove LINKTYPE from perl-piped/Makefile.PL

c72578:
- Remove line:
  `'linkext'   => {LINKTYPE => ''},`
  Since version 5.00 of MakeMaker such a line can be deleted safely.
  MakeMaker recognizes when there's nothing to be linked.
  See: https://perldoc.perl.org/ExtUtils/MakeMaker.html
- Fixes Warning:
  `LINKTYPE set to '', no longer necessary`


oetiker:
thanks

PR #943, opened 2019-01-18
Fix: Cannot rename temporary file to final file

c72578:
- Fixes the following failing tests under Windows:
  modify1, modify2, modify3, modify4, modify5
- Use CreateFileA() with FILE_SHARE_DELETE in rrd_open.c,
  so that the outfilename can later be replaced by tmpfilename using
  write_rrd() in rrd_create.c

### DIFF
--- a/src/rrd_create.c
+++ b/src/rrd_create.c
@@ -9,14 +9,14 @@
 #include <time.h>
 #include <locale.h>
 #include <math.h>
-#include <glib.h>   // will use glist and regex
+#include <glib.h>       // will use glist and regex
 
 
 
-#include <sys/types.h>      // stat()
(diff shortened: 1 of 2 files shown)


c72578:
Here is an example to test this on Windows (modified from rrdtool tests, modify1):
<pre>
rrdtool.exe create modify-test1b.rrd --start 1300000000 --step 60 DS:a:GAUGE:120:0:U  RRA:AVERAGE:0.5:1:10 RRA:AVERAGE:0.5:5:10
rrdtool.exe tune modify-test1b.rrd RRA#0:+10
</pre>

oetiker:
thanks

PR #944, opened 2019-01-19
Use _WIN32 instead of WIN32 preprocessor macro

c72578:
- _WIN32 is the recommended preprocessor macro
- Substitute occurrences of WIN32 with _WIN32 using:
<pre>
git grep -lw -e '#ifdef WIN32' -e '#ifndef WIN32' -e 'defined(WIN32)' \
-- '*.c' '*.h' '*.xs' | xargs \
sed -i -e 's/#ifdef WIN32/#ifdef _WIN32/g' \
-e 's/#ifndef WIN32/#ifndef _WIN32/g' \
-e 's/defined(WIN32)/defined(_WIN32)/g'
</pre>

### DIFF
--- a/bindings/perl-shared/RRDs.xs
+++ b/bindings/perl-shared/RRDs.xs
@@ -24,7 +24,7 @@ extern "C" {
  */
 #define VERSION_SAVED VERSION
 #undef VERSION
-#ifndef WIN32
+#ifndef _WIN32
 #include "rrd_config.h"
 #endif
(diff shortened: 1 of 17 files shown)


c72578:
Info concerning the _WIN32 preprocessor macro:
http://nadeausoftware.com/articles/2012/01/c_c_tip_how_use_compiler_predefined_macros_detect_operating_system#WindowsCygwinnonPOSIXandMinGW

Issue #956, opened 2019-02-04
etc/rrdcached-default-lsb.in missing from 1.7.1 tarball


atsampson:
`etc/rrdcached-default-lsb.in` isn't included in `rrdtool-1.7.1.tar.gz`. This causes `configure` to fail when it tries to expand the file:

```
config.status: error: cannot find input file: `etc/rrdcached-default-lsb.in'
```

I guess this needs adding to `etc/Makefile.am` so `make dist` knows about it (are there other files in `etc` that should also be listed there?).

oetiker:
oops ... thanks ... created new tarball

oetiker:
at the moment etc content is included via the toplevel Makefile.am ... but you are right this could be migrated don into this one.

c72578:
What about the following files, which were part of rrdtool-1.7.0.tar.gz and are not in rrdtool-1.7.1.tar.gz:
`etc/rrdcached.service.in`
`etc/rrdcached.socket.in`

oetiker:
now it is properly fixed :) and the tarball recreated again

atsampson:
That configures and builds successfully for me - thanks!

Running the test suite reveals that `tests/graph1.output` is missing too, though...

oetiker:
we are getting there ... slowly :)

c72578:
What about `etc/rrdcached-init-lsb`, should it also go into the tarball?
If yes, I would add it to EXTRA_DIST in `etc/Makefile.am` and also remove the trailing space after the backslash in `etc/Makefile.am:1`

oetiker:
yes please

c72578:
Fixed in v1.7.2

PR #960, opened 2019-02-05
Fixed 'No rule to make target 'etc/rrdcached.socket'

yarda:
It's probably related to the #956 and tries to fix the following compilation error:

make[1]: *** No rule to make target 'etc/rrdcached.socket', needed by 'all-am'.  Stop.
make: *** [Makefile:502: all-recursive] Error 1


oetiker:
so true

PR #961, opened 2019-02-07
add releasee tests to travis and bump version number for rerelease wi…

oetiker:
…th fixed makefiles


c72578:
Should the following be updated too in configure.ac?
NUMVERS=1.7001
->
NUMVERS=1.7002

oetiker:
hmm all these numbers should be generated from the VERSION file ...

c72578:
Yes, this would simplify things, if these numbers are generated from the VERSION file.
For consistency, the same NUMVERS should also be used in `win32/rrd_config.h`:
NUMVERS=1.72
->
NUMVERS=1.7002

c72578:
/bin/bash: ps2pdf: command not found
https://travis-ci.org/oetiker/rrdtool-1.x/builds/489975897?utm_source=github_status&utm_medium=notification

Issue #975, opened 2019-03-02
New RPN Function ROUND


TheWitness:
**Is your feature request related to a problem? Please describe.**
Presently we have CDEF functions CEIL and FLOOR, but suspiciously absent is the ROUND function.  Am I missing something?

**Describe the solution you'd like**
Just this kind of pedestrian statistical function.  If my LINEX is 1.8, make it 2.0, if it's 1.45, make it 1.

**Describe alternatives you've considered**
I tried FLOOR and CEIL, but they did just what they were supposed to do, and did not ROUND

oetiker:
:-) seems no-one has missed it yet ... but go ahead ... PR!

TheWitness:
@oetiker, good to find you well.  let me see what I can do.

TheWitness:
Done.  https://github.com/oetiker/rrdtool-1.x/pull/976

c72578:
Merged into master.
Commit: https://github.com/oetiker/rrdtool-1.x/commit/fe0c1b81b752361ce40b485f16c4dbd72d16f7df

PR #976, opened 2019-03-03
Adding Rounding Function for RRDTool

TheWitness:
I was playing with a Cacti Graph Template the other day and saw the time related averaging that is inherit in RRDTool making values that should appear integers appear as floats on the graph.  So, I'm like, hey "I'll just use a rounding function, that'll work".  Only to find that it was missing.

Love to tool been using it for almost two decades, though I don't consider myself an expert.  Here is a little contribution from me.  Nothing magical, just a rounding function.  Related to the following bug report.

https://github.com/oetiker/rrdtool-1.x/issues/975

Thank you Tobi for all you have done for the monitoring community.  Be well.


netniV:
Always a useful thing to have

netniV:
Could probably do with allowing a parameter for the number of places to round by?

TheWitness:
You can achieve the same results by first multiplying by XX rounding, and diving by XX.  So, it can be achieved without modifying this pull.  I'll defer to @oetiker and what he wants to see.

oetiker:
since neither ceil nor floor offer that precision, I would go for the simple version

oetiker:
I will release 1.7.2 prior to adding this functionality

PR #989, opened 2019-03-27
fix segfault on non-existent RRD file when using rrdcached

themylogin:
fix segfault on non-existent RRD file when using rrdcached + rrdtool xport
(like 814ca69a3329ccc88040ed184a413e4e5adf604c does for rrdtool graph)


oetiker:
thanks

PR #991, opened 2019-04-03
Use fabs() for floating point, labs() for long

c72578:
- Fixes the following clang compiler warnings:
<pre>
  rrd_rpncalc.c:999:11: warning: using integer absolute value function
  'abs' when argument is of floating point type [-Wabsolute-value]
    if (abs(percentile) > 100) {
  rrd_rpncalc.c:1103:24: warning: using integer absolute value function
  'abs' when argument is of floating point type [-Wabsolute-value]
    int idx = round(abs(idxf));
  rrd_hw.c:519:36: warning: absolute value function 'abs' given an
  argument of type 'long' but has parameter of type 'int' which may
  cause truncation of value [-Wabsolute-value]
    new_val = ((unsigned long) abs(val)) % mod;
</pre>

### DIFF
--- a/src/rrd_hw.c
+++ b/src/rrd_hw.c
@@ -516,7 +516,7 @@ static unsigned long MyMod(
     unsigned long new_val;
 
     if (val < 0)
-        new_val = ((unsigned long) abs(val)) % mod;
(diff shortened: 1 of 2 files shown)


oetiker:
thanks

PR #994, opened 2019-04-09
Update test tune1

c72578:
- Compare dump of the updated rrd file with tune1-testorg.dump,
  which is available in tests


oetiker:
thanks

PR #995, opened 2019-04-09
Cast lua_tostring return value to char *

c72578:
- Fixes the following gcc compiler warning:
<pre>
  rrdlua.c:69:21: warning: assignment discards ‘const’ qualifier from
  pointer target type [-Wdiscarded-qualifiers]
  if (!(argv[i] = lua_tostring (L, i))) {
</pre>


oetiker:
thanks

PR #1004, opened 2019-05-19
Define _POSIX_THREAD_SAFE_FUNCTIONS (MinGW-w64)

c72578:
- time.h of MinGW-w64 requires _POSIX_THREAD_SAFE_FUNCTIONS to be
  defined in order to provide ctime_r, gmtime_r or localtime_r.
- In newer versions of winpthreads (e.g. git-7.0.0.5447.a2d94c81),
  _POSIX_THREAD_SAFE_FUNCTIONS is not defined any more in
  pthread_unistd.h. Hence define it here directly.
- This is a follow-up to commit:
  https://github.com/oetiker/rrdtool-1.x/commit/603de9a

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -448,7 +448,7 @@ AC_LINK_IFELSE(
     [ AC_LINK_IFELSE(
           [AC_LANG_PROGRAM(
                 [[#ifdef __MINGW32__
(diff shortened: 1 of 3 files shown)


oetiker:
thanks

PR #1007, opened 2019-05-28
Fix missing double values in rrdinfo.

stefangluszek:
The double values of rrd info were never properly converted to the
python dictionary causing some keys (e.g. xff) to be missing.


JoakimSoderberg:
@oetiker is any official version planned that contains this version? Without this the Python bindings are unusable

PR #1038, opened 2019-08-02
Add Perl 5.22 to .travis.yml and use dist: trusty for now

c72578:
Travis has been moving from Ubuntu trusty (14.04LTS)
to xenial (16.04LTS), which has got Perl 5.22

Switch back to trusty for now, until reason for failing tests under
xenial using "make check TESTS_STYLE="valgrind-logfile" has been
investigated


c72578:
Found the reason for the failing travis build on xenial:
libtool-bin needs to be installed.

This PR is not needed any more. I will prepare a new one.

PR #1039, opened 2019-08-02
Update .travis.yml for Ubuntu xenial (16.04LTS)

c72578:
- Install libtool-bin to fix failing "valgrind-logfile" tests
- Enable the following "valgrind-logfile" tests, which are passing now
  under xenial: create-with-source-4, vformatter1
  Add also graph1 and pdp-calc1, which have been added in the meantime.
  There are only three failing tests left: rpn1, rpn2, xport1
- Add Perl 5.22, which is the version in xenial

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
+# dist: trusty
+dist: xenial
+
 language: c
 
 perl:
+  - "5.22"
   - "5.20"
   - "5.18"
   - "5.16"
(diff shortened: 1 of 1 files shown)


c72578:
Info from `- cat tests/test-suite.log` in .travis.yml:
`./functions: line 54: libtool: command not found`

c72578:
The PR is ready from my side now.

PR #1044, opened 2019-08-21
Fix compiler warnings (-Wcast-function-type)

c72578:
- Add `tree_compare_func()` and use this function instead of
  `(GCompareDataFunc) strcmp`
  Fixes the following gcc compiler warning:
  `rrd_daemon.c:4245:34: warning: cast between incompatible function`
  `types from ‘int (*)(const char *, const char *)’`
  `to ‘gint (*)(const void *, const void *, void *)’`
  `{aka ‘int (*)(const void *, const void *, void *)’}`
  `[-Wcast-function-type]`
  `cache_tree = g_tree_new_full((GCompareDataFunc) strcmp, NULL, NULL,`

- Use `g_strcmp0()` instead of `strcmp()` in `tree_compare_func()`.
  `g_strcmp0()` compares str1 and str2 like `strcmp()`. Handles NULL
  gracefully by sorting it before non-NULL strings.
  Comparing two NULL pointers returns 0.

- Cast `free_cache_item` using `(void (*)(void)`
  Fixes the following gcc compiler warning:
  `rrd_daemon.c:4246:34: warning: cast between incompatible function`
  `types from ‘void * (*)(cache_item_t *)’`
  `{aka ‘void * (*)(struct cache_item_s *)’}`
  `to ‘void (*)(void *)’ [-Wcast-function-type]`
  `(GDestroyNotify) free_cache_item);`

### DIFF
--- a/src/rrd_daemon.c
+++ b/src/rrd_daemon.c
@@ -4181,6 +4181,19 @@ static void *listen_thread_main(
     return (NULL);
 }                       /* }}} void *listen_thread_main */
 
+/* `tree_compare_func()` is used instead of `(GCompareDataFunc) strcmp` to avoid gcc compiler warning:
+ * cast between incompatible function types from ‘int (*)(const char *, const char *)’ to ‘gint (*)(const void *, const void *, void *)’
+ * {aka ‘int (*)(const void *, const void *, void *)’} [-Wcast-function-type]
+ */
+static gint tree_compare_func(
+    gconstpointer a,
+    gconstpointer b,
+    gpointer user_data)
+{
+    (void) user_data;   /* Silence -Wunused-parameter warning */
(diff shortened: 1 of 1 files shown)


c72578:
Info concerning involved functions:

`g_tree_new_full ()`
https://developer.gnome.org/glib/stable/glib-Balanced-Binary-Trees.html#g-tree-new-full

`GCompareDataFunc ()`
https://developer.gnome.org/glib/stable/glib-Doubly-Linked-Lists.html#GCompareDataFunc

`strcmp`
http://www.cplusplus.com/reference/cstring/strcmp/

`g_strcmp0 ()`
https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strcmp0

oetiker:
thanks

PR #1045, opened 2019-08-22
Fix compiler warning (-Wstringop-truncation)

c72578:
- Use `strncat()` the following way in rrd_cgi.c:
  `strncat(dest, src, dest_size - strlen(dest) - 1);`

- Fixes the following gcc compiler warning:
  `rrd_cgi.c:1407:21: warning: ‘strncat’ output truncated before`
  `terminating nul copying as many bytes from a string as its length`
  `[-Wstringop-truncation]`
  `strncat(line, tmp, tmplen);`

### DIFF
--- a/src/rrd_cgi.c
+++ b/src/rrd_cgi.c
@@ -1404,7 +1404,7 @@ static s_var **rrdcgiReadVariables(
                 if ((unsigned) length > tmplen) {
                     if ((line = (char *) realloc(line, len)) == NULL)
                         return NULL;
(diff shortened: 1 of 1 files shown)


c72578:
Info concerning `strncat()` and string truncation:
https://developers.redhat.com/blog/2018/05/24/detecting-string-truncation-with-gcc-8/
https://www.us-cert.gov/bsi/articles/knowledge/coding-practices/strncpy-and-strncat

oetiker:
thanks

PR #1046, opened 2019-08-22
Fix wbuf segfault

### LINKED ISSUE #1043: segfault in rrdcached: socket invalid
I've been dealing with rrdcached randomly crashing and had no idea why. I finally managed to build it with debug symbols (which are stripped on Arch Linux) and run it to see the problem:

```
(gdb) bt
#0  0x00007fba7f2d5c71 in __memmove_avx_unaligned_erms () at /usr/lib/libc.so.6
#1  0x0000561a6d8e5055 in wbuf_append (sock=0x561a6f10cc00, str=0x7fba7909dba0 "1863345057 Can't use 'STATS' here.\n", len=35)
    at rrd_daemon.c:725
#2  0x0000561a6d8e54f5 in send_response (sock=0x561a6f10cc00, rc=RESP_ERR, fmt=0x561a6d8f3177 "Can't use '%s' here.\n")
    at rrd_daemon.c:848
#3  0x0000561a6d8eafb7 in handle_request
    (sock=0x561a6f10cc00, now=94671532483200, buffer=0x7fba74000b20 "STATS", buffer_size=0) at rrd_daemon.c:3138
#4  0x0000561a6d8ec18c in connection_thread_main (args=0x561a6f10cc00) at rrd_daemon.c:3626
#5  0x00007fba7f46257f in start_thread () at /usr/lib/libpthread.so.0
#6  0x00007fba7f2700e3 in clone () at /usr/lib/libc.so.6

(gdb) p *(listen_socket_t*)0x561a6f10cc00
$2 = {fd = 6, addr = 0x0, family = 1, batch_start = 94671532483200, batch_cmd = 1863345057, rbuf = 0x7fba74000b20 "STATS",
  next_cmd = 6, next_read = 6, wbuf_data = 0x561a6f106d30 "\001", wbuf_size = 94671532487584, wbuf_capacity = 94671532488000,
  permissions = 16777215, socket_group = 22042, socket_permissions = 1863350320}
```

Honestly, I have no idea how to reproduce this, but my guess is that there's some logic *somewhere* which is causing this to fail. For reference, I'm running collectd's rrdcached plugin on Arch Linux. I'm not 100% sure what's causing the bug to occur but I can offer more details depending on what's needed.

I can provide a fuller core dump if needed, but I narrowed the segfault down to this thread. The segfault happens only when the socket is accessed, and it happens before a response is given to the client. My guess is that there's some case that's not being handled correctly and it's segfaulting.

If someone else can help reproduce this bug, let me know. I've tried messing with my collectd config and it doesn't seem to have made a difference.

To clarify, I'm running on the latest versions of rrdtool and collectd on Arch Linux, and I just updated my system today.

clarfon:
This ensures that wbuf_data, wbuf_size, and wbuf_capacity are all initialised to zero. Fixes #1043. 

In general, calloc should be preferred over malloc as it preserves alignment. However, as I'm not familiar enough with the code base, I'm going to simply change this one line to fix my problem.

### DIFF
--- a/src/rrd_daemon.c
+++ b/src/rrd_daemon.c
@@ -4125,7 +4125,7 @@ static void *listen_thread_main(
                 continue;
             }
 
-            client_sock = (listen_socket_t *) malloc(sizeof(listen_socket_t));
+            client_sock = (listen_socket_t *) calloc(1, sizeof(listen_socket_t));
             if (client_sock == NULL) {
                 RRDD_LOG(LOG_ERR, "listen_thread_main: malloc failed.");
                 continue;
--- a/src/rrd_daemon.c
+++ b/src/rrd_daemon.c
@@ -3775,6 +3775,9 @@ static int open_listen_socket_unix(
         return (-1);
     }
 
+    listen_fds[listen_fds_num].wbuf_data = NULL;
+    listen_fds[listen_fds_num].wbuf_size = 0;
+    listen_fds[listen_fds_num].wbuf_capacity = 0;
     listen_fds[listen_fds_num].fd = fd;
     listen_fds[listen_fds_num].family = PF_UNIX;
     listen_fds[listen_fds_num].addr = strdup(path);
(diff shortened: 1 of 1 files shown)


clarfon:
I should add, this really is the kind of bug that can be caught by linting. I don't know too much about the most common GCC flags/linting tools for C but I doubt this is the only bug of this kind in the code base.

This bug makes rrdcached unusable for me and I would greatly appreciate a patch release for this, but will understand if this is not doable any time soon. Would appreciate knowing what the status is on that so I can apply the patch locally as necessary. Thanks!

oetiker:
thanks

clarfon:
Noticed that this wasn't actually fixing my issue; found out that another bit of code had to be added to fix. These fields *should* be set to zero now.

PR #1048, opened 2019-08-31
Add missing client symbols

pallas:
A bunch of client symbols were missing from the shared library.

Aside, it is not at all clear to me how to copy rrd_client into an external project and build it, since it depends on a bunch of other files, especially since they do not share the same license.

### DIFF
--- a/src/librrd.sym
+++ b/src/librrd.sym
@@ -81,11 +81,15 @@ rrdc_fetch
 rrdc_first
 rrdc_flush
 rrdc_flush_if_daemon
+rrdc_flushall
+rrdc_flushall_if_daemon
 rrdc_forget
 rrdc_info
 rrdc_is_any_connected
 rrdc_is_connected
 rrdc_last
+rrdc_list
+rrdc_ping
 rrdc_stats_free
 rrdc_stats_get
 rrdc_update


oetiker:
afaik @octo s original idea was to have the `rrd_client.[ch]` with a more liberal license than rrdtool to make linking with other projects easier. Not sure how the other header files got in there and if they are really necessary. 

Also the rrd_client functions afaik are internal and not go be used 'client side' ... they are not used in rrdtool code as well. The missing rrdc_ entries would be ok.

octo:
ACK, my intention was to be permissive: The FSF argues that by linking with a library / shared object, you're creating a "derived work". One consequence is that if you link with a GPL licensed library, the resulting binary must also be licensed under the GPL.

I wanted non-GPL binaries be able to use `rrd_client.h`, too. `rrd_client.h` is self-contained and does not rely on other headers.

The license (MIT license) is GPL compatible, meaning you can link this module with other GPL code, too.

If this is creating problems, pleade let me know – as I said: I wanted this to be as permissive as possible.

pallas:
Ah, ok, so `rrd_client` is intended to be internal? I will switch my project to use `rrdc` instead.

As for rrd_client.[ch], I understand the intention behind the license change. If you pull those two files into another directory and attempt to compile them, you should see something like
```
rrd_client.c:47:10: fatal error: rrd_strtod.h: No such file or directory
 #include "rrd_strtod.h"
          ^~~~~~~~~~~~~~
compilation terminated.
```
The C file specifically includes rrd_strtod.h, rrd.h, rrd_tool.h, rrd_client.h, & mutex.h

oetiker:
just had a look through the code ... there are indeed a bunch of rrd_ functions used in the rrd_client code now ... they have crept in while refactoring things ... I guess there is some disentanglement required to get rrd_client.[ch] moving freely again ... PRs gladly accepted

pallas:
I modified the PR to only add the missing rrdc symbols, not the rrd_client symbols. I will take a look at making rrd_client a stand-alone file again, hopefully soon, but as a separate PR.

c72578:
Would you mind sorting librrd.sym again, using the sort command (default settings)?
`sort librrd.sym`

Then, rrdc_flushall will be after rrdc_flush_if_daemon ...

pallas:
Yes, of course. I did sort it but my default sort settings on this host appear to be ignoring underscores. Let me try changing the collation function..

PR #1049, opened 2019-09-01
Clean up tests leftover files

nirgal:
Bold try. I haven't tested.
But Debian package system is complaining about these two extra files when building several time in a row...


oetiker:
thanks

PR #1054, opened 2019-09-10
Use DefaultPlatformToolset for .vcxproj files

c72578:
- Use `$(DefaultPlatformToolset)` for the PlatformToolset
  instead of v140. This allows to use the associated PlatformToolset
  automatically for each version of Visual Studio:
  v140 for VS2015, v141 for VS2017, v142 for VS2019.
  However, a developer can easily target projects to a desired,
  different PlatformToolset.
- Set `ToolsVersion="Current"` instead of `ToolsVersion="14.0"`
- Strings were replaced in .vcxproj files the following way:
<pre>
  git grep -l 'v140' -- '*.vcxproj' | \
    xargs sed -i 's/v140/$(DefaultPlatformToolset)/g'
  git grep -l '14.0' -- '*.vcxproj' | xargs sed -i 's/14.0/Current/g'
</pre>


c72578:
The following dialogues will not appear any more, which makes opening the rrd.sln file in Visual Studio more straight forward.
VS2017:
VS2017_Review_Solution_Action
VS2019:
VS2019_Review_Solution_Action

PR #1072, opened 2019-12-30
Fix AM_CHECK_PYTHON_HEADERS macro for Python3

SuperHeron:
This detects Python3 include paths like /usr/include/python3.7m.
Taken from pygobject.


oetiker:
thanks

PR #1079, opened 2020-03-14
Fix double meaning of time 0 as uninitialized value

### LINKED ISSUE #1078: Segmentation fault in rrd_graph:print_calc when vidx is -1 on GF_VRULE
**Describe the bug**

munin-cron intermittently exits with a SIGSEGV. Upon examination of the backtrace, the fault occurs in inside `librrd.so` in [`rrd_graph:print_calc` when `im->gdes[i].gf==GF_VRULE` and `vidx==-1`](https://github.com/oetiker/rrdtool-1.x/blob/v1.7.2/src/rrd_graph.c#L2047).

**To Reproduce**

I cannot reproduce this bug on purpose (yet). With munin-cron running every five minutes, it might happen a few times a day (1-6) and then go several days without happening at all. I don't understand what the trigger is.

**Expected behavior**

Don't segfault.

**Screenshots**

With guidance, I would be glad to work on extracting more useful information from a core dump. I have not used rrdtool directly before, so it's hard for me to do better than brute force to make progress at this point.

**Desktop (please complete the following information):**

Server, actually.

 - OS: FreeBSD 12.1-RELEASE-p1 amd64
 - Browser: N/A
 - Version: rrdtool 1.7.2, perl 5.30.1, munin 2.0.56

**Smartphone (please complete the following information):**

 N/A

**Additional context**

None.

neirbowj:
Treat "VRULE:0#..." as "VRULE:1#..." because elsewhere xrule==0 is
used to mean xrule has not been set, which can lead to an array
bounds violation.

Fixes #1078

### DIFF
--- a/src/rrd_graph_helper.c
+++ b/src/rrd_graph_helper.c
@@ -1026,6 +1026,10 @@ static graph_desc_t *newGraphDescription(
                     }
                     if (gf == GF_VRULE) {
                         gdp->xrule = val;
+                        if (gdp->xrule == 0) {
+                            /* distinguish from uninitialized */
+                            gdp->xrule++;
+                        }
                     } else {
                         gdp->yrule = val;
                     }


neirbowj:
In order to permit "VRULE:0#..." it would be necessary to create an additional struct member to track whether the value in `xrule` is valid or not. Let me know if you would prefer that approach.

PR #1087, opened 2020-07-05
Add JSONTIME output to graphv data

netniV:
So, this morning I was asked a question by my Cacti colleague @browniebraun:

> May I can ask you something about RRDtool as you are more experienced with it C code?  I would like to enhance our Zoom JS code to have a hover effect showing the related data for every data point to make RRDtool graph more dynamic. Our current plans are not to switch to a different graph engine AFAIK.
> How difficult would it be to enhance the graphv option RRDtool provides to include all data points in JSONTIME format beside the blob image and the rrdgraph characteristics it already returns in a big array?

So I took a quick peek, explained what I thought would need doing, found myself tinkering, and after an hour of that here is the result.... basically I add a new element after the `image=` called `datapoints=` which is simply the output that would have been produced if `--imgformat=JSONTIME` was set.

**graphv --imgformat=jsontime**
[rrdtool-graphv-json.txt](https://github.com/oetiker/rrdtool-1.x/files/4875960/rrdtool-graphv-json.txt)
This file contains the data when exported via the JSONTIME imageformat

**graphv --imgformat=png**
[rrdtool-graphv-png.txt](https://github.com/oetiker/rrdtool-1.x/files/4875961/rrdtool-graphv-png.txt)
This file contains the data when exported as a PNG image

**graphv --imgformat=png --add-jsontime**
[rrdtool-graphv-combined.txt](https://github.com/oetiker/rrdtool-1.x/files/4875963/rrdtool-graphv-combined.txt)
This file contains the data when exported as a PNG image with the JSONTIME data added in for good measure by the `--add-jsontime` parameter.


netniV:
@oetiker we would like to see this option introduced, any feedback on it?

oetiker:
can you add a test please

netniV:
What kind of test would you like? How many ?

oetiker:
I would imagine setting TZ to something other than UTC, starting rrdtool graph with the utc option and then checking that it runs in utc mode based on some timestamp returned ....

netniV:
I will work on this and get the PR updated.

netniV:
Took me longer to write the tests than I did the patch but I think I've succeeded.  Btw, there is a test that is already failing `create-with-source-4`.  I don't believe that is anything down to what I was doing since it appears to have the following issue:

```sh
./create-with-source-4: line 87: dc: command not found
./create-with-source-4: line 87: [: -lt: unary operator expected
```

c72578:
@netniV, for `create-with-source-4` the **`dc`** command needs to be installed.

tests/Makefile.am
```diff
@@ -17,7 +17,7 @@ EXTRA_DIST = Makefile.am \
 	rrdcreate \
 	tune1-testa-mod1.dump tune1-testa-mod2.dump tune1-testorg.dump \
 	tune2-testa-mod1.dump tune2-testorg.dump \
-	valgrind-supressions dcounter1 dcounter1.output graph1.output vformatter1 rpn1.output rpn2.output \
+	valgrind-supressions dcounter1 dcounter1.output graph1.output graoup2.output vformatter1 rpn1.output rpn2.output \
```
c72578, line 20: `graoup2.output -> graph2.output`

c72578:
@netniV, did you already have a look at **`tests/graph2-valgrind.log`** after
```console
make check TESTS_STYLE="valgrind-logfile" TESTS="graph2"``
```
There are several "blocks are definitely lost" in graph2-valgrind.log

netniV:
I don't know about that that test. 

What I did do is take the xport1 test as a starting point but only used one of the DS. Could that be the reason ?  I just did the `make checks` command which passed.

netniV:
Oh and I realised the error of dc later. Shouldn't that be tested for in the build of the tests if it is required though?

c72578:
The valgrind tests are also part of the Travis-CI. See:
https://github.com/oetiker/rrdtool-1.x/blob/37cac323f9206b07e3266c8382f84f9ca8f9503f/.travis.yml#L31

Ideally a test like the new graph2 should be added there and pass. Otherwise it should be stated in the comment together with the other failing valgrind tests.
```yaml
# Disable the following, failing tests: rpn1 rpn2 xport1
# These tests are failing on Travis CI (currently Ubuntu xenial, 16.04LTS), when using valgrind-logfile
```

tests/Makefile.am
netniV: Good spot!

netniV:
valgrind.log is empty and as I am not sure what to do to fix it, I'm kinda stuck with it that way.  I have corrected the typo though.

```
./functions: line 58: libtool: command not found
cat: /usr/share/cacti/rrdtool/tests/graph2-valgrind.log.12862.tmp: No such file or directory
FAILED: (rc=127) create
FAIL graph2 (exit status: 1)
```
That was in the test suite log file.

Not sure why it thinks libtool isn't installed though:

```
apt install libtool
Reading package lists... Done
Building dependency tree
Reading state information... Done
libtool is already the newest version (2.4.6-10).
0 upgraded, 0 newly installed, 0 to remove and 2 not upgraded.
```

c72578:
@netniV, the **`libtool`** binary is distributed via the **`libtool-bin`** package. Please try:
```console
sudo apt-get install libtool-bin
```

netniV:
OK, so the functions issue has now gone away but it is still failing...

```
=========================================
   rrdtool 1.7.2: tests/test-suite.log
=========================================

# TOTAL: 1
# PASS:  0
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: graph2
============

/usr/bin/libtool: line 11733: exec: valgrind: not found
cat: /usr/share/cacti/rrdtool/tests/graph2-valgrind.log.10244.tmp: No such file or directory
FAILED: (rc=127) create
FAIL graph2 (exit status: 1)
```

netniV:
Honestly, I have no idea what I am doing here 😱 

If you can fix up the patch for whatever I've overlooked, I can bear that in mind for next time but as far as I can tell on my system it works.

This was based on the xport1, so if you deem that means it should be excluded, I have no real thoughts either way 👍

c72578:
> 
> /usr/bin/libtool: line 11733: exec: valgrind: not found

This should help:
```console
sudo apt-get install valgrind
```

netniV:
Did that still didn't understand it, gave up now.

TheWitness:
valgrind is a leak detector.

c72578:
Attached is the **`graph2-valgrind.log`** file created under Fedora 32, using
`make check TESTS_STYLE="valgrind-logfile" TESTS="graph2"`

[graph2-valgrind.zip](https://github.com/oetiker/rrdtool-1.x/files/5024195/graph2-valgrind.zip)

netniV:
Again, still don't know what I'm looking at there.  The codebase itself has not really changed, all I did was combine the use of two different functions (as you will see from the source diff).  If you did that against `graphv` for `--imgformat=PNG` and `--imgformat=JSONTIME`, does it also leak?  If so, I would say there was an underlying problem.

If not, then I have no clue what needs to be released that isn't.  Could it be the `im_free(&im)` that comes later needs a separate one for the jsontime data?  However, as that's a single point, which pulls the grinfo out of it, I don't see that as being wrong myself and the `img` variables are not released within that section of code.

tests/Makefile.am
```diff
@@ -17,7 +17,7 @@ EXTRA_DIST = Makefile.am \
 	rrdcreate \
 	tune1-testa-mod1.dump tune1-testa-mod2.dump tune1-testorg.dump \
 	tune2-testa-mod1.dump tune2-testorg.dump \
-	valgrind-supressions dcounter1 dcounter1.output graph1.output vformatter1 rpn1.output rpn2.output \
+	valgrind-supressions dcounter1 dcounter1.output graph1.output group2.output vformatter1 rpn1.output rpn2.output \
```
c72578, line 20: The typo is still not fixed (it is just a different typo since fc9027e):
`group2.output -> graph2.output`
The file is called **`graph2.output`**, otherwise `make dist` will fail.

oetiker:
thanks!

PR #1088, opened 2020-07-20
Add --utc to graph option.

jyavenard:
This is particularly useful in environment where changing the TZ environment variable is not desirable, such as in a browser script.
The data generated would be added to the rrd file while TZ=UTC, and read in the browser with --utc set.

This allows to easily calculate daily average ranging from midgnight to midnight rather than say the last 24 hours.

### DIFF
--- a/bindings/python/rrdtoolmodule.c
+++ b/bindings/python/rrdtoolmodule.c
@@ -582,6 +582,7 @@ static char _rrdtool_graph__doc__[] = "Create a graph based on one or more " \
     [-b|--base value]\n\
     [-W|--watermark string]\n\
     [-Z|--use-nan-for-all-missing-data]\n\
+    [--utc]\n\
     DEF:vname=rrdfile:ds-name:CF[:step=step][:start=time][:end=time]\n\
     CDEF:vname=RPN expression\n\
     VDEF=vname:RPN expression\n\n\
@@ -1094,7 +1095,7 @@ _rrdtool_lastupdate(PyObject *Py_UNUSED(self), PyObject *args)
             }
 
             PyDict_SetItemString(ds_dict, ds_names[i], val);
-            
+
             if (val != Py_None)
                 Py_DECREF(val);
 
@@ -1204,7 +1205,7 @@ _rrdtool_fetch_cb_wrapper(
         rrd_set_error("expected 'start' key in callback return value to be "
             "of type int");
         goto gil_release_err;
-    } else if (PyObject_RichCompareBool(tmp, tmp_min_ts, Py_EQ) || 
+    } else if (PyObject_RichCompareBool(tmp, tmp_min_ts, Py_EQ) ||
                PyObject_RichCompareBool(tmp, po_start, Py_LT)) {
         rrd_set_error("expected 'start' value in callback return dict to be "
             "equal or earlier than passed start timestamp");
--- a/doc/rrdgraph.pod
+++ b/doc/rrdgraph.pod
@@ -562,6 +562,11 @@ If one DS is missing, either because the RRD is not available or
 because it does not contain the requested DS name, just assume that we
 got empty values instead of raising a fatal error.
 
+[B<--utc>]
+
+Force the timezone to be UTC. Equivalent to doing TZ=UTC rrdtool graph .... 
+This is useful to calculate daily average ranging from midnight to midnight rather than say the last 24 hours.
+
 =head2 Data and variables
 
 B<DEF:>I<vname>B<=>I<rrdfile>B<:>I<ds-name>B<:>I<CF>[B<:step=>I<step>][B<:start=>I<time>][B<:end=>I<time>]
--- a/src/rrd_graph.c
+++ b/src/rrd_graph.c
@@ -42,12 +42,17 @@
 #include "plbasename.h"
 #endif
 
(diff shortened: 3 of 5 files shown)


jyavenard:
See #375 for history. Created a new pull request as it had been created for version 1.4; and I don't know how to update a pull request against a new branch on my side.

oetiker:
looking good

c72578:
Usage of `timegm` needs some attention under Windows.
See e.g.:
https://stackoverflow.com/questions/16647819/timegm-cross-platform
A mingw-w64 build or PR #1088 will currently fail with the following output:

```console
rrd_graph.c: In function 'find_first_time':
rrd_graph.c:52:26: warning: implicit declaration of function 'timegm'; did you mean 'time'? [-Wimplicit-function-declaration]
   52 | #define MKTIME(a,b) (b ? timegm(a) : mktime(a))
      |                          ^~~~~~
rrd_graph.c:1611:12: note: in expansion of macro 'MKTIME'
 1611 |     return MKTIME(&tm, utc);
      |            ^~~~~~
rrd_graph.c:52:26: warning: nested extern declaration of 'timegm' [-Wnested-externs]
   52 | #define MKTIME(a,b) (b ? timegm(a) : mktime(a))
      |                          ^~~~~~
rrd_graph.c:1611:12: note: in expansion of macro 'MKTIME'
 1611 |     return MKTIME(&tm, utc);
      |            ^~~~~~
...
ld.exe: .libs/librrd_la-rrd_graph.o: in function `find_first_time':
src/rrd_graph.c:1611: undefined reference to `timegm'
ld.exe: .libs/librrd_la-rrd_graph.o: in function `find_next_time':
src/rrd_graph.c:1675: undefined reference to `timegm'
collect2.exe: error: ld returned 1 exit status
```

jyavenard:
> 
> 
> can you add a test please

How do you go about testing something returning a graph?

oetiker:
you can have rrdgraph output in text format ... JSON for example ...

oetiker:
> > ...
> > librrd-8.lib(rrd_graph.obj) : error LNK2019: unresolved external symbol _timegm referenced in function _find_first_time
> 
> how do you compile on windows? is this a supported platform?

yes windows is supported and wolfgang (@c72578) is the windows master

it seems that

```
#define timegm _mkgmtime
```

might fix the problem

oetiker:
@c72578 github actions support windows ... it would be extra cool if we could have automated testing for windows too ...

jyavenard:
> 
> 
> @c72578 github actions support windows ... it would be extra cool if we could have automated testing for windows too ...

I'm not sure if this is addressed to me.

But that's really outside the scope of this change.

I've been using this patch for over 10 years so I could easily generate such graph 
https://mediaserver.avenard.org/power/search?small=1&graph=2&view=day&date=19/07/2020&hour=23&minute=59&plot=total,ch4,ext1,ext2,ext

which present data from midnight to midnight.

Someone asked me to rebase this patch on the current version which I did.

I looked at the existing test for graph, and there's just one doing very little. I feel it's unreasonable to ask for more than that.

But now this is getting too much work for which I have little interest.

So it seems I'll just have to continue building my own modified copy of rrdtool.

oetiker:
the windows testing bit is NOT adressed to you ... hence the @c72578 in front of it :)

c72578:
@jyavenard here is a modification to this PR, which allows compilation under Windows.
```diff
diff --git a/src/rrd_graph.c b/src/rrd_graph.c
index b92c5a96..2fbef664 100644
--- a/src/rrd_graph.c
+++ b/src/rrd_graph.c
@@ -49,6 +49,9 @@
 
 #include <time.h>
 #define LOCALTIME_R(a,b,c) (c ? gmtime_r(a,b) : localtime_r(a,b))
+#ifdef _WIN32
+#define timegm _mkgmtime
+#endif
 #define MKTIME(a,b) (b ? timegm(a) : mktime(a))
 
 #include <locale.h>
```

jyavenard:
> 
> 
> @jyavenard here is a modification to this PR, which allows compilation under Windows.
>  #include <time.h>
>  #define LOCALTIME_R(a,b,c) (c ? gmtime_r(a,b) : localtime_r(a,b))
> +#ifdef _WIN32
> +#define timegm _mkgmtime
> +#endif

I've already pushed similar changes yesterday

jyavenard:
Ok, after spending a couple of hours, I don't see how I can write an automated test to verify the behaviour of this. The data isn't exported with the JSON format.

Here is an example of what I would expect it to do.
I collect the power usage and store 1 minute average in a rrd database. The data is stored using UTC timestamps.
So midnight this morning, would have the time epoch time 1595376000.

Now if I wanted to present this data with the graph generated via a php called via apache; the time of midnight UTC would be displayed as 10:00AM instead of 00:00.
E.g if I wanted to display from 1595376000 to 1595462400 (23/07/2020 at 00:00AM)
You can't modify the TZ in a php script running inside apache, this would affect *all* sites running within that apache session, the logs, the content displayed, everything, and that would be bad.

So I would see this graph:
image

but I want to see:
image

note the time range displayed. 10AM vs 12AM

oetiker:
hmmm, ok I see that could prove tricky ...

c72578:
> 
> 
> @c72578 github actions support windows ... it would be extra cool if we could have automated testing for windows too ...

OK, I have opened PR #1089 now, which adds a GitHub Action for Windows

PR #1098, opened 2020-12-11
Use a separate pango fontmap per thread

nomis:
Reusing the same fontmap from multiple threads is not safe, despite the attempts to use a mutex to protect it there are other instances where it behaves unexpectedly if graphs are created from multiple threads at the same time, e.g.:

```
(process:76234): Pango-WARNING **: 11:47:25.823: failed to create cairo scaled font, expect ugly output. the offending font is 'DejaVu Sans Mono 8'
(process:76234): Pango-WARNING **: 11:47:25.823: font_face status is: no error has occurred
(process:76234): Pango-WARNING **: 11:47:25.823: scaled_font status is: invalid matrix (not invertible)
(process:76234): Pango-WARNING **: 11:47:25.823: shaping failure, expect ugly output. shape-engine='PangoFcShapeEngine', font='DejaVu Sans Mono 8', text='Apparent temp (?C)'
```

This error results in boxes instead of characters on the graph. In the worst case scenario there are memory management errors that cause GLib to complain or the process to abort.

Calling pango_cairo_font_map_get_default() multiple times from the same thread will return the same object, so this satisfies the performance expectations of reusing it within a single thread. Trying to lock a mutex for even more pango/cairo function calls is likely to decrease performance when multiple threads are used concurrently.

### DIFF
--- a/src/rrd_graph.c
+++ b/src/rrd_graph.c
@@ -485,11 +485,9 @@ int im_free(
             free(im->rendered_image);
         }
 
-        mutex_lock(im->fontmap_mutex);
         if (im->layout) {
             g_object_unref(im->layout);
         }
-        mutex_unlock(im->fontmap_mutex);
     }
 
     if (im->ylegend)
@@ -4822,8 +4820,6 @@ void rrd_graph_init(
 {
     unsigned int i;
     char     *deffont = getenv("RRD_DEFAULT_FONT");
-    static PangoFontMap *fontmap = NULL;
-    static mutex_t fontmap_mutex = MUTEX_INITIALIZER;
     PangoContext *context;
 
     /* zero the whole structure first */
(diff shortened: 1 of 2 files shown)


nomis:
Test script: [test.py.txt](https://github.com/oetiker/rrdtool-1.x/files/5680992/test.py.txt)
Test file info: [5449-1000-029AF902EE.rrd.txt](https://github.com/oetiker/rrdtool-1.x/files/5680995/5449-1000-029AF902EE.rrd.txt)

nomis:
After this is fixed, it would be helpful if Ubuntu Bionic (using 1.7.0) was updated to include these patches:
f06463ffd0c7f5e81251235987a748e18c897b1c
bd5f3ea5368f39726e4e79a7f63d6a2883766fcc
138fd097ba32246f0ff29b544e2a9541f14868a9

Compared to Ubuntu Xenial (using 1.5.5) the support for generating graphs from multiple threads under Python is awful. Lines on the graphs jump around all over the place.

There is nothing in the CHANGES file to indicate that these major threading issues in 1.7.0 are fixed in later versions to prompt anyone to upgrade.

oetiker (approved):
wow very nice

PR #1110, opened 2021-03-13
Switch to pkgconf (MSYS2, MinGW-w64)

c72578:
The MSYS2 project has switched from **`pkg-config`** to **`pkgconf`** as default.
For details, see: https://github.com/msys2/MSYS2-packages/pull/2197
Both packages work when building RRDtool for Windows using MSYS2.
- Switch to pkgconf instead of pkg-config
- Update the following files:
  `.github/workflows/ci-workflow.yml`
  `win32/README-MinGW-w64`
- This avoids the following warning in the Windows CI-build:
  `warning: removing 'mingw-w64-x86_64-pkg-config' from target list`
  `because it conflicts with 'mingw-w64-x86_64-pkgconf'`

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -17,7 +17,7 @@ jobs:
         with:
           msystem: MINGW64
           update: true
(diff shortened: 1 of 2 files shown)


oetiker:
thanks

PR #1117, opened 2021-05-26
Fix RPATH for the python bindings

### LINKED ISSUE #1116: python bindings hardcodes RPATH
**Describe the bug**
Even if configured with `--disable-rpath` it hardcodes RPATH to the python bindings.

**To Reproduce**
1. Configure with --disable-rpath
2. Build

**Expected behavior**
No RPATH in python bindings

**Desktop (please complete the following information):**
 - OS: Fedora Rawhide (f35)
 
**Additional context**
I think the problem is in the Makefile.am:
```
...
LD_RUN_PATH=$(libdir) \
$(PYTHON) ${abs_srcdir}/$@/setup.py build_ext $(PYTHON_RPATH) \
...
```
With `--disable-rpath` the PYTHON_RPATH is correctly set to empty string, but the setup-tools hardcodes the RPATH to $libdir, because there is `LD_RUN_PATH=$(libdir)`.

yarda:
Fixes #1116

Signed-off-by: Jaroslav Škarvada <jskarvad@redhat.com>

### DIFF
--- a/bindings/Makefile.am
+++ b/bindings/Makefile.am
@@ -67,7 +67,6 @@ python:
           && env \
                ABS_TOP_SRCDIR=${abs_top_srcdir} \
                ABS_TOP_BUILDDIR=${abs_top_builddir} \
-               LD_RUN_PATH=$(libdir) \
                $(PYTHON) ${abs_srcdir}/$@/setup.py build_ext $(PYTHON_RPATH) \
           && env \
                ABS_TOP_SRCDIR=${abs_top_srcdir} \
--- a/bindings/Makefile.in
+++ b/bindings/Makefile.in
@@ -763,7 +763,6 @@ python:
           && env \
                ABS_TOP_SRCDIR=${abs_top_srcdir} \
                ABS_TOP_BUILDDIR=${abs_top_builddir} \
-               LD_RUN_PATH=$(libdir) \
                $(PYTHON) ${abs_srcdir}/$@/setup.py build_ext $(PYTHON_RPATH) \
           && env \
                ABS_TOP_SRCDIR=${abs_top_srcdir} \


oetiker:
this completely disables the rpath in all cases, this means the python bindings will not work when librrd is no in the default library search patch, right ?

yarda:
> this completely disables the rpath in all cases, this means the python bindings will not work when librrd is no in the default library search patch, right ?

This is what the `--disable-rpath` should do, shouldn't it? E.g. in Fedora it's not permitted to use RPATH [1], so it is not possible to achieve it without patching.

[1] https://docs.fedoraproject.org/en-US/packaging-guidelines/#_beware_of_rpath

yarda:
> > this completely disables the rpath in all cases, this means the python bindings will not work when librrd is no in the default library search patch, right ?
> 
> This is what the `--disable-rpath` should do, shouldn't it? E.g. in Fedora it's not permitted to use RPATH [1], so it is not possible to achieve it without patching.
> 
> [1] https://docs.fedoraproject.org/en-US/packaging-guidelines/#_beware_of_rpath

Also Fedora is trying to enforce no RPATH:
https://fedoraproject.org/wiki/Changes/Broken_RPATH_will_fail_rpmbuild

The feature is called "broken RPATH...", but it also includes explicit standard RPATHs like e.g. $libdir.
It seems openSUSE already enforces no RPATH, Debian discourages use of it. Python upstream is discussing about it. So I think there should be configure switch for it. There is currently `--disable-rpath`, but it doesn't disable all RPATHs.

c72578:
In **`bindings/Makefile.am`** we have
```
if ENABLE_RPATH
PYTHON_RPATH=--rpath=$(libdir)
PERL_RPATH=RPATH=$(libdir)
endif
```

> Instead of specifying the -rpath to the linker, the environment variable LD_RUN_PATH can be set to the same effect [1].
>

So, there is some redundancy and it makes sense to remove `LD_RUN_PATH=$(libdir)`, which is currently not inside `if ENABLE_RPATH`.

[1] https://en.wikipedia.org/wiki/Rpath#The_role_of_GNU_ld

oetiker:
ah, cool ... so the question is whether PYTHON_RPATH is take into consideration ...

PR #1139, opened 2021-11-08
Draft: Add support for TUNE command in rrdcached

Frazew:
# Summary

## Context 

The `TUNE` command is not supported by rrdcached (see issue https://github.com/librenms/librenms/issues/4356). Running `rrdtool tune` with the `-D` argument is only used for FLUSH and still requires rrdcached and rrdtool to be using a shared filesystem.

## Solution

This PR aims to fix this by implementing the TUNE command in rrdcached and modifying rrdtool to properly run it on the remote server instead of locally when the `-D` argument is supplied.

# Testing

## Context

This PR originates from the fact that LibreNMS supports using a remote rrdcached for storage. Because rrdtool currently lacks the TUNE function over a remote rrdcached, issues arise when one tries to tune the ports [as described in the documentation](https://docs.librenms.org/Extensions/RRDTune/) (see also the issue mentioned above).

## Sample infrastructure

This patch was introduced within a production infrastructure that can be basically summarized like this:

```
+----------------------------------------------+
|                     librenms infrastructure  |
|                                              |
|                                              |
|                                              |
| +----------+      +-----------+              |
| | frontend |      | rrdcached |              |
| | 10.1.0.3 +------+ 10.1.0.42 |              |
| +----------+      +---+-+-+---+              |
|                       | | |                  |
|                       | | |                  |
|                       | | |                  |
|                    +--+-+-+--+               |
|                    | pollers |               |
|                    +---------+               |
|                                              |
+----------------------------------------------+
```

## Validating

We confirm that the patch is functional with the following steps, checked on `sample_hostname` with a port whose speed had recently been upgraded.

### Checking the RRD file

```shell
$ rrdtool info --daemon 10.1.0.42:42217 sample_hostname/port-id996775.rrd | head -30
[...]
ds[INOCTETS].max = 1.2500000000e+1
```

### Tuning using LibreNMS

```shell
$ cd /opt/librenms/ && cat config.d/rrdcached.php
<?php
$config['rrdcached'] = "10.1.0.42:42217";
$config['rrdtool_version'] = '1.7.2';
$ ./scripts/tune_port.php -h sample_hostname -p port-channel4094.3
Found hostname sample_hostname.......
Tuning port port-channel4094.3.......
$ exit
```

### Checking the RRD file again

```shell
$ rrdtool info --daemon 10.1.0.42:42217 sample_hostname/port-id996775.rrd | head -30
[...]
ds[INOCTETS].max = 2.5000000000e+10
```

# Notes

- **This patch is being tested within a production infrastructure and the PR description is updated as we get more insights**
- I am by no means a professional C developer, reviewers might want to take this into account!

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -46,14 +46,14 @@ jobs:
         include:
           - os: windows-2019
             triplet: x64-windows
-            # https://github.com/microsoft/vcpkg/commit/b361c2eefa3966cb7cec45275aff32e90430aaa6
-            vcpkgCommitId: 'b361c2eefa3966cb7cec45275aff32e90430aaa6'
+            # https://github.com/microsoft/vcpkg/commit/70033dbb31527fb3e69654731f540f59c87787f9
+            vcpkgCommitId: '70033dbb31527fb3e69654731f540f59c87787f9'
             vcpkgPackages: 'cairo expat fontconfig freetype gettext glib libpng libxml2 pango pcre zlib'
             configuration: 'x64'
             nmake_configuration: 'USE_64BIT=1'
           - os: windows-2019
             triplet: x86-windows
-            vcpkgCommitId: 'b361c2eefa3966cb7cec45275aff32e90430aaa6'
+            vcpkgCommitId: '70033dbb31527fb3e69654731f540f59c87787f9'
             vcpkgPackages: 'cairo expat fontconfig freetype gettext glib libpng libxml2 pango pcre zlib'
             configuration: 'x86'
             nmake_configuration: ''
--- a/.github/workflows/release-windows.yml
+++ b/.github/workflows/release-windows.yml
@@ -25,14 +25,14 @@ jobs:
         include:
           - os: windows-2019
             triplet: x64-windows
-            # https://github.com/microsoft/vcpkg/commit/b361c2eefa3966cb7cec45275aff32e90430aaa6
-            vcpkgCommitId: 'b361c2eefa3966cb7cec45275aff32e90430aaa6'
+            # https://github.com/microsoft/vcpkg/commit/70033dbb31527fb3e69654731f540f59c87787f9
+            vcpkgCommitId: '70033dbb31527fb3e69654731f540f59c87787f9'
             vcpkgPackages: 'cairo expat fontconfig freetype gettext glib libpng libxml2 pango pcre zlib'
             configuration: 'x64'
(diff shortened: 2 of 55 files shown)


landy2005:
@Frazew hi, can you make ability for use rrdcached with `dump` and `restore` commands?
(`dump` already should work, but really not: #1141)

Frazew:
I've edited the description to reflect the tests that were conducted which tend towards indicating that the patch should be working as intended.

oetiker:
> I've edited the description to reflect the tests that were conducted which tend towards indicating that the patch should be working as intended.

Hi @Frazew, thank you a lot for your contribution. Great to hear that this is working for you in produciton already. Could you please update the tests in the tests folder to exercise the new functionality ... in that way the testing would not be limited to you saying so in the PR.

thanks tobi

Frazew:
Hi @oetiker! I've had to somewhat significantly modify the way the tests are conducted in order for them to be actually useful here.

The problem was that communication between rrdtool and rrdcached was always tested using a unix domain socket, which leads to a fundamentally different behavior in the `get_path`  implementation. Also, because both rrdtool and rrdcached were pointing to the same directory, commands that would fallback to the local implementation (like the tune command) worked like a charm while they should not have.

I fixed this by implementing the `rrdcached-tcp` and `rrdcached-unix` test suites. The `rrdcached-unix` behaves mostly the same way as before while `rrdcached-tcp` launches rrdcached on a tcp socket (which disables the unix domain socket logic) and sets `$RRDTOOL` to a function (in `tests/functions`) which fixes the calls that are not actually implemented in rrdcached.

One great way to ensure that the new test suite behaves as intended is to apply its changes on the current `master` branch and try to run the `rrdcached-tcp` test suite. All calls to `TUNE` fail unless the workaround is enabled [here](https://github.com/oetiker/rrdtool-1.x/pull/1139/files#diff-d4121dd95ec2cb496a48c5e61fced4a9909237bb2e01b0c5db7a158ca761a141R175).

Hopefully this makes the PR more complete! This should also facilitate the tests if one decides to implement `dump` and `restore` as reported in #1141 and #1142

oetiker:
great! will have a look!

Frazew:
fyi I've just pushed a quick change that should fix the failing windows builds (failing PCRE download from vcpkg)

c72578:
> fyi I've just pushed a quick change that should fix the failing windows builds (failing PCRE download from vcpkg)

In the meantime, this has been fixed in master (#1148) and merged into this PR.

oetiker (approved):
thank you!

PR #1153, opened 2022-01-20
Fix for a race condition in journal_write

usaleem-ix:
There is a race condition in journal_write() where journal_lock is
being acquired after checking whether journal_fh is NULL or not.
journal_fh is a static file handle that can be set to NULL by any
other thread, while current thread is blocked by
pthread_mutex_lock(). This commit fixes this race condition.

Signed-off-by: Umer Saleem <usaleem@ixsystems.com>

### DIFF
--- a/src/rrd_daemon.c
+++ b/src/rrd_daemon.c
@@ -3412,10 +3412,12 @@ static int journal_write(
 {                       /* {{{ */
     int       chars;
 
-    if (journal_fh == NULL)
-        return 0;
-
     pthread_mutex_lock(&journal_lock);
+    if (journal_fh == NULL) {
+        pthread_mutex_unlock(&journal_lock);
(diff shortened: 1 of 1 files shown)


oetiker:
thank you for the fix!

oetiker (approved):

usaleem-ix:
You are welcome. Do I need to look into CodeQL / Analyze (cpp) (pull_request) ?

oetiker:
> You are welcome. Do I need to look into CodeQL / Analyze (cpp) (pull_request) ?

No, you don't have to. But if you like to, by all means do! Once we get CodeQL happy for the whole project, I would make it a requirement to be happy for each PR.

usaleem-ix:
Sure, thank you.

PR #1154, opened 2022-01-26
clear LC_ALL when setting LC_TIME in rpn2 test

mwhudson:
This test was failing when LC_ALL was set to C.UTF-8 and using the C.UTF-8 locale which is part of glibc 2.35, which differs in a very minor way to that that has been used by Debian/Ubuntu for years. The tests sets LC_TIME to C, presumably to try to isolate itself from such things, but that does not have any effect if LC_ALL is set in the build environment.


oetiker (approved):
thanks!

PR #1160, opened 2022-03-14
Update version to 1.8.0 for MSVC builds

c72578:
Includes version and copyright year updates to files in the win32
subdirectory after running the `rrdtool-release` script.


oetiker:
hmmm wondering ... could we make it so that it configures this at build time, based on the VERSION file ?

PR #1161, opened 2022-03-15
rrdtool-release: Create NUMVERS from VERSION file

c72578:
In the script **`rrdtool-release`**, `NUMVERS` has been read from the
`configure` file so far, which may not be up-to-date.
Generate `NUMVERS` from the `VERSION` file instead. This ensures
that only one source is used for the version - the `VERSION` file.
Use the same perl command for creating `NUMVERS`, as in `configure.ac`

### DIFF
--- a/rrdtool-release
+++ b/rrdtool-release
@@ -2,7 +2,7 @@
 # shellcheck disable=SC2086,SC2046,SC2029
 set -e
 VERSION=$(cat VERSION)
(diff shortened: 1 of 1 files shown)


c72578:
This is, how `NUMVERS` is generated in `configure.ac`:
https://github.com/oetiker/rrdtool-1.x/blob/c1e083e810cc1f3a07bccc9bfbfcd36604bb9084/configure.ac#L653

oetiker:
thanks

PR #1162, opened 2022-03-20
acinclude.m4: Include <stdlib.h> when using exit

ryandesign:
Fixes:

```
conftest.c:124:33: error: implicitly declaring library function 'exit' with type 'void (int) __attribute__((noreturn))'
```


oetiker:
thanks

Issue #1163, opened 2022-03-20
Remove or update generated files in repository


ryandesign:
**Describe the bug**
The repository contains generated autotools files—configure, several Makefile.in, the conftools directory, etc.—but they were not updated for the recent 1.8.0 release. If a user clones the repository or downloads an [automatically generated tarball](https://github.com/oetiker/rrdtool-1.x/archive/refs/tags/v1.8.0.tar.gz) of 1.8.0 and configures and builds, the program will misidentify itself as version 1.7.2.

The files in the [release tarball](https://github.com/oetiker/rrdtool-1.x/releases/download/v1.8.0/rrdtool-1.8.0.tar.gz) were updated properly, so users downloading the release tarball will not have this problem.

I suggest either removing the generated files from the repository (most projects do this) or updating them to the ones that are in the release tarball.

**To Reproduce**
1. Download https://github.com/oetiker/rrdtool-1.x/archive/refs/tags/v1.8.0.tar.gz
2. `configure`, `make`, `sudo make install` as usual
3. Run `rrdtool --version`

**Expected behavior**
```
$ rrdtool --version
RRDtool 1.8.0  Copyright by Tobias Oetiker <tobi@oetiker.ch>
               Compiled Mar 20 2022 16:37:57

Usage: rrdtool [options] command command_options
Valid commands: create, update, updatev, graph, graphv,  dump, restore,
		last, lastupdate, first, info, list, fetch, tune,
		resize, xport, flushcached

RRDtool is distributed under the Terms of the GNU General
Public License Version 2. (www.gnu.org/copyleft/gpl.html)

For more information read the RRD manpages
```

**Actual behavior**
```
$ rrdtool --version
RRDtool 1.7.2  Copyright by Tobias Oetiker <tobi@oetiker.ch>
               Compiled Mar 20 2022 11:05:37

Usage: rrdtool [options] command command_options
Valid commands: create, update, updatev, graph, graphv,  dump, restore,
		last, lastupdate, first, info, list, fetch, tune,
		resize, xport, flushcached

RRDtool is distributed under the Terms of the GNU General
Public License Version 2. (www.gnu.org/copyleft/gpl.html)

For more information read the RRD manpages
```

**Desktop (please complete the following information):**
 - OS: macOS 10.15.7

PR #1164, opened 2022-03-22
Update Version and Copyright year in git files

c72578:
This commit is based on running the first part of the `rrdtool-release`
script (before `./bootstrap`).

- Part of #1163


c72578:
Something for a separate PR:
- `configure`, `Makefile.in` files
These could be removed, as they are generated during `bootstrap` from `configure.ac` and `Makefile.am`
- Update of the download link of the tarball

oetiker (approved):

PR #1102, opened 2021-01-11
Reproducibility improvement

kleinik:
RRDtool binaries include build date and time, no matter if SOURCE_DATE_EPOCH environment variable was set or not. Following changes were made in the context of [Reproducible Builds](https://reproducible-builds.org/) initiative.

PR #1165, opened 2022-03-23
Fix BUILD_DATE in rrdtool help output

c72578:
- This is a followup to #1102
- Fixes segfault when running "rrdtool --help"
- Change `DATE_FMT` to the same date format as the `__DATE__` macro [1]:
  mmm dd yyyy

[1] https://gcc.gnu.org/onlinedocs/cpp/Standard-Predefined-Macros.html

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -695,7 +695,7 @@ AC_MSG_RESULT(${COMP_PERL:-No Perl Modules will be built})
 
 # Use reproducible build date and time
 if test "$SOURCE_DATE_EPOCH"; then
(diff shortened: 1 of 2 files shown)


oetiker (approved):

c72578:
Additional information: This is what happens without this fix.
- Windows, mingw32 or mingw64 build:
```console
$ ./rrdtool.exe
Segmentation fault
```
- Linux (build time is null):
```console
$ rdtool
RRDtool 1.8.0  Copyright by Tobias Oetiker <tobi@oetiker.ch>
               Compiled 22 Mar 2022 00:00:00 (null)
```

PR #1224, opened 2023-04-04
doc/rrdbuild: Add 'dc' to Debian build dependencies

diederikdehaas:
Without `dc` the `create-with-source-4` test fails.


diederikdehaas:
It turns out this is the last dependency needed. After installing ``dc`` I could build all the Debian rrd-tool packages, so not only ``make check``, but the full build succeeded.

oetiker (approved):

Issue #1219, opened 2023-04-04
Transparent Areas/Stacks are being overlayed, resulting on odd lines showing


thurban:
**Describe the bug**

**To Reproduce**
Charts with transparent colors show odd lines when using Area/Stack. The different areas are being overlayed on each other which cause the part where they overlap to be with a stronger color.

**Expected behavior**
The different areas should not overlap each other.

**Screenshots**
Chart showing Issue ( 0.2 value )
![RRDTool_Issue](https://user-images.githubusercontent.com/17292964/229726855-60329463-a9f2-4b57-9eb6-50dd9a0a2fa3.png)
Chart showing fixed ( 0 value )
![RRDTool_Fixed](https://user-images.githubusercontent.com/17292964/229726857-08e62b40-c752-491b-9cf3-169ffc5815ce.png)

**Desktop (please complete the following information):**
rrdtool 1.6.0, 1.7.0, 1.7.1, 1.8.0 

**Additional context**

The overlap is caused by the "0.2" in the following code from [ rrd_graph.c:](https://github.com/oetiker/rrdtool-1.x/blob/master/src/rrd_graph.c#L4297 )
```
if (im->slopemode == 0) {
                            backY[++idxI] = ybase - 0.2;
                            backX[idxI] = ii + im->xorigin - 1;
                            foreY[idxI] = ytop + 0.2;
                            foreX[idxI] = ii + im->xorigin - 1;
                        }
                        backY[++idxI] = ybase - 0.2;
                        backX[idxI] = ii + im->xorigin;
                        foreY[idxI] = ytop + 0.2;
                        foreX[idxI] = ii + im->xorigin;
```

Then setting the 0.2 to "0" ( nod adding/removing anything from ytop/ybase) the graph show up just fine

oetiker:
### Triage check (master @ 17d1f842)

**Status:** Likely fixed

The root cause was the `±0.2` pixel overdraw applied to `ybase` and `ytop` in `src/rrd_graph.c` (around line 4297 at the time of the report), which caused adjacent transparent AREA/STACK fills to overlap by 0.4 pixels, producing visible darker seams. This overdraw was removed in commit d8923c5c ("Fix for overlapping areas in transparent mode (#1226)"), merged in May 2023. The fix ships in v1.9.0 and all later releases (v1.10.x); current master contains the corrected code with exact `ybase`/`ytop` coordinates and no overdraw.

oetiker:
[...]

PR #1226, opened 2023-04-19
Fix for overlapping areas in transparent mode

thurban:
Fix for  Transparent Areas/Stacks are being overlayed, resulting in odd lines showing #1219


oetiker (approved):

oetiker:
can you please also update the CHANGES file, describing the effect of the change

thurban:
HI,

I've updated the CHANGES file with new entry in the Bugfixes section

oetiker (approved):

PR #1242, opened 2023-12-25
Constify argv, fix warnings.

hramrach:
rrd has no business modifying the string pointed to by passed agrv, and
    as far as gcc can see it does indeed not modify them because it compiles
    with const argv.
    
    This fixes warnings when passing const strings into rrd, and avoids the
    need to duplicate all strings in tcl bindings.
    
    This fixes warnings like these:
    [    3s] prog/sensord/rrd.c: In function 'rrdInit':
    [    3s] prog/sensord/rrd.c:302:40: warning: cast discards 'const' qualifier from pointer target type [-Wcast-qual]
    [    3s]   302 |                 ret = rrd_create(argc, (char**) argv);
    [    3s]       |                                        ^
    [    3s] prog/sensord/rrd.c: In function 'rrdUpdate':
    [    3s] prog/sensord/rrd.c:458:42: warning: cast discards 'const' qualifier from pointer target type [-Wcast-qual]
    [    3s]   458 |                 if ((ret = rrd_update(3, (char **) /* WEAK */ argv))) {
    [    3s]       |                                          ^

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -41,6 +41,9 @@ Features
 
 * Add (remote) dump support to rrdcached <Tobias Hintze>
 
+* Constify argv argument to library functions
+  Without this users of library function that want to pass in const strings need to duplicate them to avoid compiler warnings
+
 RRDtool 1.8.0 - 2022-03-13
 ==========================
 
--- a/bindings/lua/rrdlua.c
+++ b/bindings/lua/rrdlua.c
@@ -37,8 +37,8 @@
 extern void rrd_freemem(void *mem);
 
 extern int luaopen_rrd (lua_State * L);
-typedef int (*RRD_FUNCTION)(int, char **);
-typedef rrd_info_t *(RRD_FUNCTION_V)(int, char **);
+typedef int (*RRD_FUNCTION)(int, const char **);
+typedef rrd_info_t *(RRD_FUNCTION_V)(int, const char **);
 
 /**********************************************************/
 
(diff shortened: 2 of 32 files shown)


oetiker:
@hramrach thanks for the patch! Please update the CHANGES file as well, then I think we are ready to merge.

c72578:
Under Windows msys2 there is an additional test failing now:
`FAIL: pdp-calc1`
https://github.com/oetiker/rrdtool-1.x/actions/runs/7397250681/job/20133555409#step:5:1832

We should have a look.

c72578:
> Under Windows msys2 there is an additional test failing now: `FAIL: pdp-calc1` https://github.com/oetiker/rrdtool-1.x/actions/runs/7397250681/job/20133555409#step:5:1832
> 
> We should have a look.

OK, I tested with current master under msys2 and `FAIL: pdp-calc1` has appeared there too.
So, it is unrelated to this PR (#1242) and will be cared about separately ...

c72578 (dismissed):

CHANGES
```diff
@@ -41,6 +41,9 @@ Features
 
 * Add (remote) dump support to rrdcached <Tobias Hintze>
 
+* Constify argv argument to libary functions
```
c72578, line 44: libary->library

hramrach:
Is it really a new failure? I see other failure there.

Also does it use some language bindings? I don't see any connection there.

The pure C changes are only type casts, I initially included more changes but then reverted them to make it clear that nothing changed there.

c72578 (approved):

oetiker (approved):

Issue #1223, opened 2023-04-04
Several compiler warnings during build (``make check``)


diederikdehaas:
**Describe the bug**
When running ``make check VERBOSE=1`` (on amd64), I'm seeing several compiler warnings of the following categories:  
- incompatible-pointer-types
- format-truncation=
- stringop-truncation

**To Reproduce**
1. clone repo; cd repo-dir; ``./bootstrap``; ``./configure``
2. Run ``make check VERBOSE=1``
3. See compiler issue several warnings (See output below)

**Expected behavior**
A build output without compiler warnings.

**Screenshots**
```sh
diederik@bagend:~/dev/github/rrdtool-1.x$ make check VERBOSE=1
Making check in po
make[1]: Entering directory '/home/diederik/dev/github/rrdtool-1.x/po'
make rrdtool.pot-update
make[2]: Entering directory '/home/diederik/dev/github/rrdtool-1.x/po'
sed -e '/^#/d' remove-potcdate.sin > t-remove-potcdate.sed
mv t-remove-potcdate.sed remove-potcdate.sed
if LC_ALL=C grep 'GNU rrdtool' ../* 2>/dev/null | grep -v 'libtool:' >/dev/null; then \
  package_gnu='GNU '; \
else \
  package_gnu=''; \
fi; \
if test -n 'https://github.com/oetiker/rrdtool-1.x/issues' || test '' = '@'PACKAGE_BUGREPORT'@'; then \
  msgid_bugs_address='https://github.com/oetiker/rrdtool-1.x/issues'; \
else \
  msgid_bugs_address=''; \
fi; \
case `/usr/bin/xgettext --version | sed 1q | sed -e 's,^[^0-9]*,,'` in \
  '' | 0.[0-9] | 0.[0-9].* | 0.1[0-5] | 0.1[0-5].* | 0.16 | 0.16.[0-1]*) \
    /usr/bin/xgettext --default-domain=rrdtool --directory=.. \
      --add-comments=TRANSLATORS: --keyword=_ --keyword=N_  \
      --files-from=./POTFILES.in \
      --copyright-holder='Tobias Oetiker' \
      --msgid-bugs-address="$msgid_bugs_address" \
    ;; \
  *) \
    /usr/bin/xgettext --default-domain=rrdtool --directory=.. \
      --add-comments=TRANSLATORS: --keyword=_ --keyword=N_  \
      --files-from=./POTFILES.in \
[...]

PR #1244, opened 2024-01-06
Fix string truncation warnings related to PATH_MAX

hramrach:
There are a number of places where rrdtool combines multiple PATH_MAX sized strings into one.

PATH_MAX is a constant that tends to work in practice a lot of the time but may not reflect the real capabilities of the system in real time.

In place of on-stack buffers of PATH_MAX size allocate memory dynamically. Initialize the pointers to NULL so they can be all freed unconditionally on exit.

Fixes: #1223

Note: this is not tested, and the changes are substantial. If this looks like a generally good way to do it it warrants running the code with valgrind or similar.

Also it is not clear that asprintf implementation is needed. configure tests for asprintf presence but a windows implementation is provided.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -13,6 +13,7 @@ Bugfixes
 * acinclude.m4: Include <stdlib.h> when using exit <ryandesign>
 * rrdtool-release: Create NUMVERS from VERSION file <c72578>
 * Avoids leaking of file descriptors in multi threaded programs by @ensc
+* Avoids potential unterminated string because of fixed PATH_MAX buffer
 
 Features
 --------
--- a/src/rrd_daemon.c
+++ b/src/rrd_daemon.c
@@ -112,6 +112,7 @@
 
 #include "rrd_strtod.h"
 #include "compat-cloexec.h"
+#include "rrd_snprintf.h"
 
 #include <glib.h>
 /* }}} */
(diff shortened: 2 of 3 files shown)


c72578:
> Also it is not clear that asprintf implementation is needed. configure tests for asprintf presence but a windows implementation is provided.

Concerning Windows, MSYS2 MinGW-w64 provides `asprintf`:
`checking for asprintf... yes`

And for the MSVC builds (librrd-8.vcxproj) there are:
[win32/asprintf.c](https://github.com/oetiker/rrdtool-1.x/blob/master/win32/asprintf.c)
[win32/asprintf.h](https://github.com/oetiker/rrdtool-1.x/blob/master/win32/asprintf.h)

oetiker:
the build-test-linux does run valgrind ... not sure if the tests are hitting path areas ... maybe have a look?


this is how to run the tests on the commandline

```
        make check TESTS_STYLE="valgrind-logfile" TESTS="modify1 modify2 modify3 modify4 modify5 tune1 tune2 graph1 rrdcreate dump-restore create-with-source-1 create-with-source-2 create-with-source-3 create-with-source-4 create-with-source-and-mapping-1 create-from-template-1 dcounter1 vformatter1 list1 pdp-calc1"

```

hramrach:
And there is whole printf implementation so there should not be a problem with any platform lacking asprinf.

Yes, I see that the tests run valgrind, and the coverage tool points out that some of the error cases are not tested.

oetiker (approved):

oetiker (changes requested):
please add to the CHANGES file!

oetiker (approved):

PR #1246, opened 2024-01-07
Drop win32-specific asprintf implementation

hramrach:
asprintf is implemented by rrd_snprintf anyway, the win32-specific implementation is redundant.


c72578:
Very nice cleanup.

hramrach:
It also tests that a platform without native asprintf is supported.

c72578 (dismissed):

oetiker (approved):

src/rrd_info.c
```diff
@@ -8,9 +8,7 @@
 #include "rrd_rpncalc.h"
 #include "rrd_client.h"
 #include <stdarg.h>
-#ifdef _MSC_VER
-#include "asprintf.h"   /* for vasprintf() here */
-#endif
+#include "rrd_snprintf.h"   /* for vasprintf() here */
```
c72578, line 11: Please check:
`/* for vasprintf() here */`
However, in **`rrd_info.c`** `vasprintf` is only used inside `#ifdef HAVE_VASPRINTF`

hramrach: Interestingly, it does build on Windows where the vasprintf was supposedly missing.

Will have another look.

hramrach: There is also the thing that asprintf implemented using the platform-specific vsnprintf uses locale-specific formatting while rrd_asprintf does not.

oetiker: NOT using locale specific formatting was one of the reasons for going with our private implementation as locale specific formatting wrecks havoc on applications trying to parse the outpus

oetiker: well in any case it would be good if the windows version worked the same as the unix version ... human readable output is primarely in the graphics

hramrach: rrd_list uses printf only for string arguments so there the implementation should be interchangeable.

rrd_info uses asprintf for string arguments and integer arguments without thousands separator so this again should produce same output regardless of asprintf implementation.

There is one use of e conversion with plain printf that would be affected by locale which is not changed in this PR.

oetiker (dismissed):

c72578 (approved):
Even more cleanup. LGTM.

oetiker (approved):

oetiker:
thank you @hramrach !

PR #1248, opened 2024-01-19
cloexec support on Windows

hramrach:
Microsoft documentation suggests that the 'N' mode flag to fopen is equivalent to POSIX O_CLOEXEC.


hramrach:
Does not work

PR #1255, opened 2024-07-12
Fix extra reference of parameters of rrd_fetch_dbi_{long,double}

jamborm:
Functions rrd_fetch_dbi_long and rrd_fetch_dbi_double currently take the first parameter of type "bi_result result *" even though that is already a pointer and the use of that parameter suggests the extra indirection is not wanted.  What is more, the caller passes just "bi_result result" to the corresponding actual arguments which results in compile errors with GCC 14 because it now does not accept incompatible pointer types by default.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -14,6 +14,7 @@ Bugfixes
 * rrdtool-release: Create NUMVERS from VERSION file <c72578>
 * Avoids leaking of file descriptors in multi threaded programs by @ensc
 * Avoids potential unterminated string because of fixed PATH_MAX buffer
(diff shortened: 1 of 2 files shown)


c72578:
@jamborm Could you please post the original GCC 14 warnings/errors here concerning `rrd_fetch_dbi_double` and `rrd_fetch_dbi_long`?

jamborm:
Sure: 
[rrdtool-errors.log](https://github.com/user-attachments/files/16232076/rrdtool-errors.log)

PR #1260, opened 2024-07-29
Update further files to version 1.9.0

c72578:
Includes version and copyright year updates to files after running the
`rrdtool-release` script.


oetiker (approved):
oops you had me worried there ... but I think in the tar attached to the release, all is setup properly ... uff ... thanks for updating master accordingly.

PR #1266, opened 2024-10-12
illumos defines _MAX macros in limits.h

hadfl:


c72578 (commented):
The only `_MAX` in **`rrd_list.c`** is `PATH_MAX`. So, this commit is about missing `PATH_MAX` under illumos?

`PATH_MAX` is also required in **`rrd_daemon.c`** [1].
Therefore, could you please consider handling missing `#include <limits.h>` in case of `__illumos__` somewhere else, more globally?
e.g. in **`src/rrd_config_bottom.h`**

https://github.com/oetiker/rrdtool-1.x/blob/39e01491e58b9aec790de8f29530cb97489391d5/src/rrd_config_bottom.h#L30

[1]
```sh
git grep PATH_MAX
CHANGES:* Avoids potential unterminated string because of fixed PATH_MAX buffer
src/rrd_config_bottom.h:# ifdef PATH_MAX
src/rrd_config_bottom.h:#  define MAXPATH PATH_MAX
src/rrd_daemon.c:        if ((end_ptr - start_ptr + strlen(fullpath) + 1) >= PATH_MAX) {
src/rrd_list.c:         if (strlen(dirname) + strlen(entry->d_name) + 1 >= PATH_MAX) {
```

hadfl:
Thank you for your review. You are right, it's much better to put it into `src/rrd_config_bottom.h`.

src/rrd_config_bottom.h
```diff
@@ -27,6 +27,12 @@
 #ifdef HAVE_SYS_PARAM_H
 # include <sys/param.h>
 #endif
+
+/* illumos defines *_MAX macros in limits.h */
+#ifdef __illumos__
+#include <limits.h>
```
c72578, line 33:
```suggestion
# include <limits.h>
```

hadfl: Done

oetiker (approved):

oetiker:
thanks

PR #1267, opened 2024-11-13
rrdtool-python: Prefer Python3 over Python2.3

pticon:
Python2.3 is deprecated and therefore, the Python bindings should use Python3 as default.


TheWitness (approved):
Looks good.

pticon:
@oetiker Is there any comments on it ?

oetiker (approved):

PR #1283, opened 2025-03-27
Mark ruby bindings as ractor safe

LevitatingBusinessMan:
My other PR isn't merged yet but I would also like for you to look at this one.

Ruby has a new method for concurrency called [Ractor](https://docs.ruby-lang.org/en/master/Ractor.htm)s.

For C bindings to be used inside a Ractor they need to be marked Ractor-safe:

> If an extension desires to be marked as Ractor-safe the extension should call rb_ext_ractor_safe(true) at the Init_ function for the extension, and all defined methods will be marked as Ractor-safe.

https://docs.ruby-lang.org/en/master/extension_rdoc.html#label-Appendix+F.+Ractor+support

I think all of the methods present in the bindings are inherently thread safe.
By marking them Ractor-safe ruby programs can generate graphs concurrently.

See also https://github.com/ruby/ruby/pull/3824 for more information concerning this method.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -3,7 +3,7 @@ RRDtool - master ...
 ====================
 Bugfixes
 --------
-
+* Mark the Ruby bindings as Ractor safe @LevitatingBusinessMan
 * Fix Compatiblity with TCL-9 @yselkowitz
 * Correctly link Ruby Bindings @LevitatingBusinessMan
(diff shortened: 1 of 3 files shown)


oetiker (dismissed):

c72578:
@LevitatingBusinessMan Please update the `CHANGES` file too.

c72578 (approved):

PR #1309, opened 2026-03-11
fix: add pkg-config fallback for Perl and Ruby bindings

somethingwithproof:
## Summary
- Perl and Ruby bindings fail to build standalone (outside autotools) because they hardcode in-tree paths
- Adds three-way detection: Windows (unchanged), in-tree (ABS_TOP_BUILDDIR), standalone (pkg-config)
- `die`/`abort` on missing librrd in standalone path instead of silent `exit 0`

## Test plan
- [ ] `perl Makefile.PL` with ABS_TOP_BUILDDIR set (in-tree build)
- [ ] `perl Makefile.PL` with librrd installed via pkg-config (standalone)
- [ ] `ruby extconf.rb` with both paths

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -11,6 +11,8 @@ Bugfixes
 * Fix compilation on illumos @hadfl
 * Python2.3 is deprecated and therefore, the Python bindings should use Python3 as default @pticon
 * Fix issue where RRDtool detects a LINE or AREA with a constant numeric value as being exportable
+* Add pkg-config fallback for Perl and Ruby bindings when building standalone @somethingwithproof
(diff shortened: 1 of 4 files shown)


somethingwithproof:
The Ruby build failure is because extconf.rb checks `ENV['ABS_TOP_BUILDDIR']` but the Makefile.am only passes it as a make variable to the subsequent `$(MAKE)` call, not to the `$(RUBY) extconf.rb` invocation.

The Python binding already handles this correctly with `env ABS_TOP_BUILDDIR=... $(PYTHON) setup.py`. Fix incoming -- same pattern for Ruby.

oetiker (approved):

PR #1311, opened 2026-03-11
fix: escape control characters in JSON xport output (RFC 7159)

somethingwithproof:
## Summary
- `rrdtool xport --json` emits raw control characters (U+0000-U+001F) in legend strings
- RFC 7159 section 7 requires these be escaped as `\uXXXX`
- Rewrites `escapeJSON()` with forward-tracked safe boundary for correct truncation

## Test plan
- [ ] Legend containing tab, newline, null byte produces valid JSON
- [ ] Long legends are truncated at safe escape boundaries
- [ ] Normal ASCII legends unchanged


somethingwithproof:
Codecov flags are from the new escape paths not having dedicated edge-case tests. The xport2 test covers the happy path. Could add a test with control chars in DS names if that'd help, but the build itself is clean.

somethingwithproof:
Codecov flags are from the new escape paths not having dedicated edge-case tests. The xport2 test covers the happy path. Could add a test with control chars in DS names if that'd help, but the build itself is clean.

oetiker (dismissed):

oetiker:
@somethingwithproof please update the CHANGES file!

somethingwithproof:
CHANGES already included in the PR.

somethingwithproof:
CHANGES entry was already included in the PR (see commit). Requesting a re-review when you have a chance. Thanks!

PR #1312, opened 2026-03-11
fix: safety checks in xport, graph, graph_helper, and daemon

somethingwithproof:
## Summary
Addresses 15 safety issues across 4 files:

**rrd_xport.c**: NULL check after malloc in escapeJSON, cleanup loop unsigned wraparound fix, step==0 division guard, int cast removal on row/col counts, OOB read guard when now < gdes.start

**rrd_graph.c**: realloc NULL check (temp pointer pattern), AREA malloc NULL checks for foreY/foreX/backY/backX, VDEF_PERCENT zero-steps guard, VDEF_PERCENTNAN zero-nancount guard

**rrd_graph_helper.c**: memmove +1 for null terminator in legend_shift

**rrd_daemon.c**: gmtime_r thread safety, vsnprintf (removed unsafe vsprintf fallback), handle_request_tune argv bounds check, handle_request_create av[] bounds check with MAX_CREATE_AV constant, check_pidfile pid_str zero-init, argc<=0 rejection

## Test plan
- [ ] `rrdtool xport` with step=0 returns error (not SIGFPE)
- [ ] `rrdtool graph` with empty VDEF dataset returns DNAN (not SIGFPE)
- [ ] `rrdcached` rejects CREATE with 128+ DS/RRA definitions
- [ ] `rrdcached` rejects TUNE with 0 arguments


somethingwithproof:
The modify/tune test failures are from the handle_request_tune argc hardening. The original code used atoi() on the next field after the filename, which silently returned 0 for non-numeric strings like `RRA#0:+10` -- that's the normal case since the rrdtool client doesn't send an argument count for tune. The strtol replacement rejects non-numeric input, breaking the protocol.

Fix restores the fallback: if the field isn't a valid number, treat it as the first argument and collect the rest dynamically. Strict parsing still applies when an explicit count is provided.

oetiker:
@somethingwithproof something in the tests with xport3 is failing on linux ...

somethingwithproof:
I will spin up a docker container and take a look.

somethingwithproof:
Investigated the xport3 Linux failure. The likely cause is the bounds guard I added at rrd_xport.c:424:

```c
if (im->gdes[vidx].step > 0 &&
    now >= im->gdes[vidx].start) {
```

The original code only checked `step > 0`. Adding `now >= start` was intended to prevent a negative index when `now < start`, but on Linux the RRD start alignment can place the first export row at exactly `start` (or slightly before after modular alignment at line 397: `*start = im->start - im->start % (*step)`). If the aligned start lands before `gdes[vidx].start`, the first data point becomes DNAN instead of a real value.

The xport3 test checks JSON legend escaping, not data values, so the DNAN itself should not affect the assertions. However, if the DNAN causes the output row count or structure to change (e.g., an empty data array on a single-step export), the python `json.loads()` validation could fail on a structural difference.

Fix: relax the guard to use `>=` with the aligned start rather than per-row `now`:

```c
if (im->gdes[vidx].step > 0) {
    long idx = (long)floor((double)(now - im->gdes[vidx].start) / im->gdes[vidx].step);
    if (idx >= 0) {
        chosen_idx = idx * im->gdes[vidx].ds_cnt + im->gdes[vidx].ds;
        (*dstptr++) = im->gdes[vidx].data[chosen_idx];
    } else {
        (*dstptr++) = DNAN;
    }
}
```

This preserves the safety (no negative index) without changing the output for rows at or after `start`. Will push the fix shortly.

somethingwithproof:
Pushed fix. Two changes in the latest commit:

1. **xport data fill**: replaced `now >= im->gdes[vidx].start` with an index-based check (`idx >= 0` after computing the signed floor index). This preserves negative-index safety without changing output when the aligned export start equals the data source start.

2. **xport3 truncation bug**: the escaped legend was formatted through `snprintf(buf, 256, ...)` which truncated the 1020-char escaped output mid-`\uXXXX`, producing invalid JSON. Fixed by writing the legend directly to the string buffer via `addToBuffer()`.

Verified on Linux (Docker Ubuntu 22.04): xport3 now passes. The 8 remaining failures (modify1-5, tune1-2, create-with-source-4) are all rrdcached-dependent and unrelated to this PR.

c72578:
The test `xport3` fails under Windows.
`FAILED: (rc=1) json legend carriage-return escaped`

See:
https://github.com/oetiker/rrdtool-1.x/actions/runs/24330382176/job/72149045845#step:5:1737

Test log file from an msys2 MINGW64 build:
[xport3.log](https://github.com/user-attachments/files/26908914/xport3.log)

The following screenshot of  `xport3.log` at the failing part shows that there is `crend` inside xport3.log instead of `cr\rend`:

crend

However, `xport3_cr.json` correctly contains `"cr\rend"`
[xport3_cr.json](https://github.com/user-attachments/files/27163977/xport3_cr.json)

c72578:
In the meantime I could have a closer look at the failing test `xport3` under Windows and MINGW64.
As it is not an issue of #1312 here, a separate PR with a fix has been prepared: #1315.

c72578 (approved):

oetiker:
thanks

PR #1313, opened 2026-03-23
A warning message appears when fontconfig is initialized in FreeBSD.

je3kmz:
Since the following changes were made to fontconfig, the message "Fontconfig warning: too many callers of FcObjectFini()" is being output to standard output almost every time munin-monitoring plots graphs.
https://gitlab.freedesktop.org/fontconfig/fontconfig/-/commit/dbbb46207e243d07d2e4c4d35547fc7521c9c62f

I don't know where the problem lies, but applying the following patch will make the warning disappear.
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=293213


oetiker:
please add a CHANGES entry!

oetiker (dismissed):

oetiker:
@je3kmz there seems to be a problem with fcinit in the windows setup ... you have to add conditionals around the call

je3kmz:
I can't verify this on Windows. :-(

c72578:
@je3kmz The check for fontconfig and `FcInit` is disabled in configure.ac

https://github.com/oetiker/rrdtool-1.x/blob/f6e83697a352849cca6bdd42a82a43dd76d67fa1/configure.ac#L650

`FcInit` should only be called, if it has been checked for, is available and included.

oetiker:
> I can't verify this on Windows. :-(

but you can enable the test in configure and wrap your change in a conditional, then the code will compile :)

oetiker:
ping ?

je3kmz:
I apologize for the delayed report.

Even after modifying configure.ac, running autoconf -f -i to generate the configuration file, and installing with the generated configuration file,
the same message still appears on standard output.

This issue occurs in a FreeBSD environment.

I have not been able to confirm it on other operating systems.

Applying the aforementioned patch to src/rrd_graph.c prevents this issue from occurring.

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=293213
This site is where I deal with FreeBSD-related issues. The patch for src/rrd_graph.c posted there is a bit old, but
the patch I pulled request also prevents this issue in a FreeBSD environment.

je3kmz:
https://cgit.freebsd.org/ports/commit/?id=c83dec7d604e44c61921425a38a00fa135e1f076

This patch works correctly, so my problem has been resolved.

PR #1315, opened 2026-04-29
Fix test xport3 under Windows msys2

c72578:
Use bash nameref in function `xport_json()`.

- Fixes error: `FAILED: (rc=1) json legend carriage-return escaped`
- Using nameref instead of command substitution preserves `\r` in test
  `Carriage return (0x0d) → \r`


c72578:
The PR depends on #1312.
Example of a successful `xport3` test under Windows (msys2, MINGW64):
[xport3.log](https://github.com/user-attachments/files/27215319/xport3.log)

oetiker (approved):

PR #1332, opened 2026-05-26
Fix heap buffer overflow in checkUnusedValues() (alt. to #1329)

oetiker:
## Summary

Counter-proposal to #1329 (kept that PR's commit message attribution).

@ppisar diagnosed a real 1-byte heap buffer overflow in `src/rrd_graph_helper.c:checkUnusedValues()` — glibc fortified `strcat()` SIGABRT in RRD-Simple-1.44 tests after upgrading to 1.10.0. PR #1329 fixes the overflow with a one-line `len = 1` change, but that change orphans the trailing-`:` strip at line 153, so the colon now survives in the error message (which @ppisar's own report shows: `"Unused Arguments \" ... CEST\r:\""`).

Root cause: the original function used the under-sized `len` *both* to allocate the buffer (`len - 1` byte short, hence the overflow) *and* to find the trailing `:` for the strip (`res[len - 1] = 0`). The bug and the strip were silently coupled — fixing one breaks the other.

## Approach

Rather than fix the indexing twice (`len = 1` *and* `res[len - 2] = 0`), refactor so the strip is gone:

- Track `used = strlen(res)` explicitly.
- Size each (re)allocation to exactly `used + sep + kvlen + 1`.
- Prepend `:` before all entries except the first.
- Use `memcpy` at the known offset rather than `strncat` (also avoids the per-iteration `strlen` walk).
- Preserve the realloc-OOM behaviour (return whatever was built).

Result string is identical to the *intended* behaviour of the original (e.g. `"alpha:beta"`, no trailing colon), and the buffer sizing and the content no longer have to agree by accident.

## Test plan

- [x] Function compiles with `-Wall -Werror` (the project's CI flags).
- [x] Standalone driver under `-fsanitize=address`, six cases:
  - empty list (no unused) → NULL
  - one unused → `"foo"`
  - two unused → `"foo=1:bar=2"`
  - mixed flagged/unflagged → `"alpha:beta"`
  - the timestamp-with-CR string from @ppisar's bug report → no overflow, NUL-terminated
  - all flagged → NULL
- [x] No ASan errors, no leaks.

The other failure @ppisar mentions — `Unused Arguments` for a `COMMENT` containing a literal `\r` — is a separate parsing bug, not addressed here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -3,6 +3,15 @@ RRDtool - master ...
 ====================
 Bugfixes
 --------
+* Fix a 1-byte heap buffer overflow in `checkUnusedValues()` that
+  triggered `*** buffer overflow detected ***` SIGABRT under glibc
+  fortified `strcat()` when reporting unused graph arguments. The
+  function under-allocated by one byte and silently relied on that
+  overflow to make its trailing `:` strip land on the right index.
+  Refactored to size buffers correctly and prepend the separator
+  rather than build-then-strip, so the two no longer have to agree.
+  Reported and original fix attempt in PR #1329 by @ppisar,
+  refactored fix by @oetiker
 * Pad the Perl `$RRDs::VERSION` / `$RRDp::VERSION` numeric encoding so
   two-digit minor releases compare monotonically. The numeric version
   now uses three-digit zero-padded minor and patch fields, e.g.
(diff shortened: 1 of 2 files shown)


andyrtr:
Applying this PR on top of any of the recent 1.10.x releases breaks monitorix creating the first kernel png for me:

> ERROR: while graphing /srv/http/monitorix//imgs/system1.1week.png: Unused Arguments " 8h 37m\c" in command : COMMENT:system uptime: 8h 37m\c

Is this expected behavior?

hazzl:
This still seems to be present in 1.11.0. I observe the same problem as descibed in #1356

PR #1352, opened 2026-05-28
rrd_xport: support use-nan-for-all-missing-data

### LINKED ISSUE #1350: xport: support -Z|--use-nan-for-all-missing-data
**Is your feature request related to a problem? Please describe.**
While **graph** supports `-Z|--use-nan-for-all-missing-data`, **xport** (as documented) doesn't

**Describe the solution you'd like**
It would be great, if the very same parameter would also be available for **xport**

**Describe alternatives you've considered**
I can of course look up all available data sources before running an export query, but for many use-cases use-nan is way more convenient.

**Additional context**
Just in case you're wondering why I'm requesting this feature: I'm working on a larger OSS project, which provides the basic infrastructure for distributed polling, with SNMP currently being the first and most prominent feature. Metrics are stored on remote polling nodes, while the related inventory is centralized. The Metrics feature is based on rrdtool, trying to abstract away some of it's complexities. Installation instructions are (intentionally) still missing, as I'm trying to avoid support overhead before tagging a first stable release. However, all of this is pretty solid so far, and running on semi-productive systems since quite a while.

Said all this, I'm  using **xport** in combination with **graph**, showing graphs as provided by rrdgraph in your web browser, while still allowing the user to get related details (e.g. current min/avg/max bits/s) for the currently mouse-hovered point in time - and that's where xport comes into play. Templates are being stored as rrdgraph_data/rrdgraph_graph combinations, currently vendor-shipped, but in future they should be customizable by users. That's why I wrote a related [parsing library](https://github.com/im-edge/rrdgraph), which also allows me to tweak graphs at will, a template editor is in the making, and I'm using this to transform graph instructions into an xport command.

My problem is, that not all monitored devices ship all required data sources. My RRD files are created only with effectively shipped data sources. Where required, the system dynamically extends them on-the-fly, deferring updates for single files, while they're being modified on disk. But this means, that for example a file containing Packet statistics (unicast, multicast etc) might have less data sources than a similar file related to another device. This means, that the following xport...

```xport --start 1779872712 -m 1138 --end 1779959112 --json
  DEF:def_average_ifInUcastPkts=c8/c8c8e9f2201e427eade8f1c3e270b747.rrd:ifInUcastPkts:AVERAGE
  DEF:def_average_ifInNUcastPkts=c8/c8c8e9f2201e427eade8f1c3e270b747.rrd:ifInNUcastPkts:AVERAGE
  DEF:def_average_ifOutUcastPkts=c8/c8c8e9f2201e427eade8f1c3e270b747.rrd:ifOutUcastPkts:AVERAGE
  DEF:def_average_ifOutNUcastPkts=c8/c8c8e9f2201e427eade8f1c3e270b747.rrd:ifOutNUcastPkts:AVERAGE
  XPORT:def_average_ifInUcastPkts:'/s Unicast packets inbound' ...
```

...triggers this error:

```ERROR: No DS called 'ifInNUcastPkts' in 'c8/c8c8e9f2201e427eade8f1c3e270b747.rrd'```

Now my DB contains all related data source definitions, and I have implemented methods allowing me to strip unavailable data sources AND related calculations, but I'd like to avoid that extra round-trip to the DB. As I'm also providing a related data source for Grafana, it would be very convenient to have such a simple "if there's nothing, ship nothing" switch available.

Related screenshots, for those who are curious: https://github.com/im-edge

### PR DESCRIPTION
**Pull request by Thomas-Gelf**
After a quick build & test, it seems that this suffices to fix #1350

### CODE REVIEW

**Comment by Thomas-Gelf** [comment]
I cloned tests/xport1 to xport4 and shortened it. The output is as expected, BUT gives `0.0000000e+00` for the very first row instead of null. As I didn't change anything there, I assume that this discrepancy already existed for rrdgraph. Now I'm unsure whether I should tweak my assumption (adjusting xport4.json.output to fit the effective export), or whether this needs further investigation. Any guidance would be greatly appreciated!

```
./tests/xport4
OK: create
OK: update 10
OK: update 20
OK: update 30
OK: update 40
OK: update 50
OK: update all
--- -	2026-05-28 17:33:16.941270016 +0200
+++ /usr/local/src/rrdtool-1.x/tests/xport4.json.output	2026-05-28 17:32:08.707983680 +0200
@@ -10,7 +10,7 @@
           ]
      },
   "data": [
-    [ null, 0.0000000e+00, null ],
+    [ null, null, null ],
     [ -5.4883500e-03, null, 3.6845819e-02 ],
     [ 0.0000000e+00, null, 3.6293297e-02 ]
   ]
FAILED: (rc=1) xport json
```

oetiker:
I would suggest to digg into the code to figure out why its working the way it does ... the rest of the testsuite seems to be fine ...

PR #1377, opened 2026-08-06
Fix right axis percentage logic and rotate right axis label

blappm:
As discussed ...

### DIFF
--- a/src/rrd_graph.c
+++ b/src/rrd_graph.c
@@ -2633,21 +2633,53 @@ int draw_horizontal_grid(
                                     (double) im->ygrid_scale.labfact;
 
                                 if (MaxY < 10 || fabs(right_step) < 1) {
-                                    snprintf(graph_label_right,
-                                             sizeof graph_label_right,
-                                             "%5.1f %s", sval,
-                                             second_axis_symb);
+                                    if (im->right_axis_percentage) {
+                                        if (sval >= 0 && sval <= 105) {
+                                            snprintf(graph_label_right,
+                                                     sizeof graph_label_right,
+                                                     "%5.1f %s", sval,
+                                                     second_axis_symb);
+                                        } else {
+                                            graph_label_right[0] = '\0';
+                                        }
+                                    } else {
+                                        snprintf(graph_label_right,
+                                                 sizeof graph_label_right,
+                                                 "%5.1f %s", sval,
+                                                 second_axis_symb);
+                                    }
                                 } else {
-                                    snprintf(graph_label_right,
-                                             sizeof graph_label_right,
-                                             "%5.0f %s", sval,
-                                             second_axis_symb);
+                                    if (im->right_axis_percentage) {
+                                        if (sval >= 0 && sval <= 105) {
+                                            snprintf(graph_label_right,
+                                                     sizeof graph_label_right,
+                                                     "%5.1f %s", sval,
+                                                     second_axis_symb);
+                                        } else {
+                                            graph_label_right[0] = '\0';
+                                        }
+                                    } else {
+                                        snprintf(graph_label_right,
+                                                 sizeof graph_label_right,
+                                                 "%5.0f %s", sval,
+                                                 second_axis_symb);
+                                    }
                                 }
                             }
                         } else {
-                            snprintf(graph_label_right,
-                                     sizeof graph_label_right,
-                                     im->second_axis_format, sval, "");
+                            if (im->right_axis_percentage) {
+                                if (sval >= 0 && sval <= 105) {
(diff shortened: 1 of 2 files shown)


oetiker:
Hi Martin — thanks for pushing this. Before we get into the code, I'd like to back up and talk about intent, because I think the patch is solving a symptom.

**What's the actual goal?**

If I read this right, the complaint is: you have a right axis that represents a percentage, and because the right axis is derived from the *left* axis' gridline positions (just `scale`/`shift` applied to them), you get labels like `130.0` or `-20.0` that make no sense for a percentage. Is that it — or is there something more?

I ask because `--right-axis-percentage` doesn't make anything a percentage. It hides labels outside a hardcoded `0 … 105`. That's a range filter wearing a percentage costume, and the name will mislead the next person who reads the docs. Two more honest framings:

- `--right-axis-range min:max` — same effect, no magic constants, works for any unit, and can suppress the *tick* as well as the text (as written you get bare unlabelled gridlines above the cutoff, which looks like a rendering bug).
- Fix the root cause: give the right axis its own grid computation instead of borrowing the left one. Bigger job, but it's the real answer, and it would also fix the long-standing weirdness where right-axis labels land on ugly values.

I'd rather land the first than the second if the second is too much work — but I'd rather not land a `--right-axis-percentage` flag, because it's an option we have to support forever and it doesn't generalise. Also, where does `105` come from?

**The legend angle**

The `RRDGRAPH_YLEGEND_ANGLE + 180` in `grid_paint()` changes the right-axis label orientation for *every* existing user, unconditionally, with no way back. `RRDGRAPH_YLEGEND_ANGLE` is a configure-time setting precisely so people can choose, and hardcoding an offset on top of it defeats that.

But you've put your finger on a real gap: the angle shouldn't be a compile-time constant at all, and the two axes shouldn't be forced to share one value. So instead of the `+180` I'd like two runtime options:

```
--vertical-label-angle <degrees>      # left, defaults to RRDGRAPH_YLEGEND_ANGLE
--right-axis-label-angle <degrees>    # right, defaults to RRDGRAPH_YLEGEND_ANGLE
```

Configure variable stays the default for both, so nobody's existing graphs change, and you get what you're after with `--right-axis-label-angle 270`. That's self-contained and uncontroversial — I'll just implement it here, it's less work than having you rebase. Then drop the angle hunk from this PR and it stays focused on the axis-labelling question.

**A real bug in the range check**

In the autoscale branch the test runs *after* `sval /= second_axis_magfact`, and the SI symbol is still appended afterwards. So the comparison is against a rescaled number, not the value the user sees:

- right axis 0…1 (a fraction), `auto_scale` picks milli → `sval` becomes 0…1000 → **every** label suppressed.
- right axis 0…100000, magfact k → `sval` becomes 0…100 → **everything** kept.

It only does what you want when magfact happens to be 1.

**Shape of the diff**

Whatever we end up calling the option, this doesn't need four copies of the same `snprintf`. One guard around the existing block does it:

```c
if (im->right_axis_percentage && (sval < 0 || sval > 105))
    graph_label_right[0] = '\0';
else
    /* existing code, untouched */
```

~5 lines instead of 49/11, no duplication to drift apart later, and it also covers the `TIMESTAMP` and `DURATION` formatters, which the current patch leaves out.

**Also missing:** `doc/rrdgraph.pod`, a `CHANGES` line ending in `@blappm`, and ideally a test under `tests/`.

So — tell me what you're actually trying to draw and let's settle the option shape first; I'd rather not have you write more code against a design we haven't agreed on.

oetiker:
Martin, I went ahead and implemented the two pieces we can agree on without further discussion: #1378.

It adds `--vertical-label-angle` and `--right-axis-label-angle`, both defaulting to `RRDGRAPH_YLEGEND_ANGLE`, so nobody's existing graphs move and you get your flipped right hand legend with `--right-axis-label-angle 270`. It also adds `--right-axis-range min:max` as the general form of the label suppression, with either end optional.

Please drop the angle hunk from this PR. Whether the range option covers your case, or whether you still need something percentage specific on top of it, is the part I'd still like to hear from you on.

One correction to what I wrote above: I claimed the range option could suppress the tick as well as the label. It can't. The mark at the right edge is the stub of the major grid line, which belongs to the left axis, so cutting it would damage the left axis too. #1378 therefore suppresses only the label, and a clipped right axis is left with unlabelled grid lines above or below its range. Worth knowing before you look at the output and think something is broken.

The other thing I ran into while doing it: the space reserved next to the graph for the legend is a fixed two font widths regardless of angle, so anything other than 90 or 270 can overlap its surroundings. That predates all of this, and I left it alone rather than widening the scope, but it is the reason the angle options are documented as being for upright labels.

PR #1383, opened 2026-09-10
Enable xport without graph support

Reinerost:
`rrdtool xport` uses the graph data processing and parser, but does not
require Cairo/Pango rendering. It already initializes the graph context
with `IMAGE_INIT_NO_CAIRO`.

This change keeps the required graph data processing code available when
RRD graph support is disabled, while guarding rendering-only code and
types with `HAVE_RRD_GRAPH`.

This allows `rrdtool xport` to remain available when configured with:

    ./configure --disable-rrd_graph

The normal graph-enabled build remains unchanged.

Tested with:
- normal graph-enabled build
- `--disable-rrd_graph` build
- RRD create/update
- XPORT XML output
- XPORT JSON output

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -54,13 +54,13 @@ RRD_C_FILES =		\
 	rrd_tune.c	\
 	rrd_list.c
 
+RRD_C_FILES += rrd_graph.c    \
+        rrd_graph_helper.c    \
+        rrd_xport.c
 
 if BUILD_RRDGRAPH
-RRD_C_FILES += rrd_graph.c	\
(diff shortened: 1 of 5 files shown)


hnyman:
One use case is embedded devices like home routers, where the graphing tool dependencies create quite much extra size compared to available flash sizes. Being able to use rrdtool to keep the dabase and to format the necessary data series with DEF,CDEF,VDEF and then pass the data series via xport json to (small) javascript (or other graphing) tools could make sense. OpenWrt is still using the really ancient 1.0.50 due to its small size, but this approach might offer a route to modern rrdtool versions.

